PR #9, opened 2019-02-25
Feat/style/enforce

Kaneraz:


Kaneraz:
A lot of playing around here, but I'm happy with where we wound up. The build will fail if we don't run code cleanup, and we can later break these scripts out to use as an optional git precommit hook.

PR #11, opened 2019-02-27
Chore/ci/pipelines

Kaneraz:
Some more pipeline changes. We run build, release, and code cleanup as separate jobs and move the yml to a build directory. We can expand by making a local build scripts and template files if we need some reuse in the pipeline yml.


Kaneraz:
This should be the configuration needed to push a tagged version to nuget.

PR #32, opened 2019-07-09
Feat/retry/add database

Kaneraz:
Use polly to handle retries and add retries add retries when getting the availability database on a secondary.


Kaneraz:
The issue wasn't with the retry logic, but that the primary wasn't being added. I still like the way polly works though. I may come back to start using it for lazy refresh and retries.

PR #34, opened 2019-12-12
chore(nuget): Add copyright/licensing nuget metadata

kevelairn:
I was looking at some other nuget packages, and there are some additional metadata fields that could be included.


Kaneraz:
Thanks!

PR #39, opened 2020-08-10
feat(cleanup): update nuget packages, cleanup code

alexirion10:
bump to .NET Core 2.2
update Nuget packages
various code cleanup


src/SmoFacade/Listener.cs
```diff
@@ -1,137 +1,134 @@
...
-       * We initially connect an availability group instance by way of a listener name. This creates a different connection and
-       * SMO server object from connecting directly to the primary instance. Having different SMO server objects results in the
-       * data cached by SMO getting out of sync between the two and forces us to refresh the objects regularly. This should
-       * prevent us from having to do the refreshes by only using SMO objects that connect directly to the server. The initial
-       * connection to find those server names though is done through the listener and that instance is thrown away in this
-       * constructor so it won't be used again.
-       */
-
-      var connectionStringBuilder = new SqlConnectionStringBuilder(connectionString);
-      connectionStringBuilder.InitialCatalog = "master";
```
Kaneraz, line 55: This is an implementation detail of how the facade uses the connection string given to connect to secondaries and shouldn't be made the responsibility of the caller. If the database name is given in the connection string it will fail to connect to the secondaries.

src/SmoFacade/Server.cs
```diff
@@ -1,209 +1,223 @@
...
+        if(loginProperties.PasswordHash != null)
+          login.Create(loginProperties.PasswordHash, LoginCreateOptions.IsHashed);
+        else if(loginProperties.Password != null)
+          login.Create(loginProperties.Password);
+        else
+          throw new ArgumentException("Password or hash was not supplied for sql login.");
+      }
+      else {
+        login.Create();
+      }
```
Kaneraz, line 208: While creating the login object requires the _server the rest of this is logic pertaining to the login, so I liked it being in the login class.

src/SmoFacade/Listener.cs
```diff
@@ -1,137 +1,134 @@
...
+  {
+    /// <summary>
+    ///   The primary instance at the time of construction.
+    /// </summary>
+    Server Primary { get; set; }
+
+    /// <summary>
+    ///   The secondary instances at the time of construction.
+    /// </summary>
+    IList<Server> Secondaries { get; set; }
```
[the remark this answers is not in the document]
alexirion10: ya... that was the point of changing it!  On https://github.com/alexirion10/AgDatabaseMove/blob/feat/cleanup/simplify/src/SmoFacade/Listener.cs#L71-L73 fixes the previous code copying all of the values.
Perhaps the better change is to make the IList<Server> private and having the Getter return IEnumerable

src/SmoFacade/Listener.cs
alexirion10: I was unsure why this was here.
> If the database name is given in the connection string it will fail to connect to the secondaries.

Actually? Why?

src/SmoFacade/Server.cs
alexirion10: The point of this was to get rid of the indirect functions returning new objects via `ConstructLogin`.
Agreed. Somewhere we'll have to get the private `_server` member. So that'd have to be exposed to put this in the Login class.

src/SmoFacade/AvailabilityGroup.cs
```diff
@@ -1,72 +1,73 @@
-﻿namespace AgDatabaseMove.SmoFacade
-{
-  using System;
-  using System.Collections.Generic;
-  using System.Linq;
-  using Exceptions;
-  using Smo = Microsoft.SqlServer.Management.Smo;
```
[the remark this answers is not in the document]
alexirion10: this was done automatically by ReSharper code cleanup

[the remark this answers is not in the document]
Kaneraz: We'll need to update the config in the repo then so running on your machine matches mine and the build.

src/SmoFacade/Listener.cs
Kaneraz: The secondary nodes don't have to be readable. If you attempt to connect to a database on a secondary that is marked as non-readable the connection will fail. Our use case in particular will have non readable secondaries.

alexirion10:
@Kaneraz - "only those with write access to this repository can merge PR's"
Can you merge this yourself or grant me write access?

Kaneraz:
@alexirion10 you should have access now. Give it another try.

PR #40, opened 2020-08-18
feat(http): url regex

alexirion10:
WIP: not tested yet


src/BackupChain.cs
```diff
@@ -4,13 +4,13 @@ namespace AgDatabaseMove
   using System.Collections.Generic;
   using System.IO;
   using System.Linq;
-  using Exceptions;
-  using SmoFacade;
+  using Microsoft.SqlServer.Management.Smo;
```
Kaneraz, line 7: smo should not be exposed from the Facade

tests/AgDatabaseMove.Unit/FileToolsTest.cs
```diff
@@ -0,0 +1,58 @@
...
+      new object[] { "https://a.diff" },
+      new object[] { "https://1/2/3/4/5/a.diff" }
+    };
+
+    public static IEnumerable<object[]> NonUrlFileExamples => new List<object[]> {
+      new object[] { @"c:\hello\a.bak" },
+      new object[] { @"\\abc\hello/a.bak" },
+      new object[] { "http://hello/a.bak" },
+      new object[] { "https://storage-account.blob.core.windows.net/container" },
+      new object[] { "https://storage-account.blob.core.windows.net/container/file.bad" }
```
Kaneraz, line 23: Aren't all of these valid URL backup possibilities?

alexirion10: None of them are
1. `c:\hello\a.bak` windows path not https
2. `\\abc\hello/a.bak` DFS path not https
3. `http://hello/a.bak` htttp not https
4. `https://storage-account.blob.core.windows.net/container` directory, not a file
5. `https://storage-account.blob.core.windows.net/container/file.bad` invalid extension

src/SmoFacade/Server.cs
```diff
@@ -104,32 +103,30 @@ public Database Database(string dbName)
...
-        if(defaultFileLocations != null) {
-          restore.RelocateFiles.Clear();
-          foreach(var file in restore.ReadFileList(_server).AsEnumerable()) {
-            var physicalName = (string)file["PhysicalName"];
-            var fileName = Path.GetFileName(physicalName) ??
-                           throw new InvalidBackupException($"Physical name in backup is incomplete: {physicalName}");
-
-            if(fileRelocation != null)
-              fileName = fileRelocation(fileName);
+      if(defaultFileLocations != null) {
```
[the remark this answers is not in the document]
alexirion10: The reason is that the `FileRelocator` func is getting called on each iteration of `foreach(var backup in backupOrder)`.
You'd only need to relocate files once.
I need to test this, but at arm's length, there isn't any comment or obvious reasoning for doing this inside the for loop

src/BackupMetadata.cs
```diff
@@ -2,6 +2,52 @@ namespace AgDatabaseMove
...
+    public enum BackupType
+    {
+      Full, // bak  / D
+      Diff, // diff / I
+      Log // trn  / L
+    }
+
+    public static bool IsUrl(string path)
+    {
+      return Regex.IsMatch(path, @"(https:\/)(\/[a-z0-9\.\-]+)+\.(bak|trn|full|diff)");
```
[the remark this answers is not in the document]
alexirion10: Why not?  It's just a way to validate that we are only using SQL Server backup files and not just any file extension. Azure blob storage (s3 as well) isn't limited to specific file types.  I can get rid of the extension, but it's really not hurting anything if it contains an exhaustive list

src/BackupChain.cs
alexirion10: This isn't in the Facade and it's used for `DeviceType.Url` and `DeviceType.File`

src/SmoFacade/Server.cs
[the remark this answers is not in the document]
Kaneraz: I can't recall the reason I'd done it that way, and I admit I should have left a comment for future me. Please revert it if it's not required or change that has been investigated in detail.

src/BackupMetadata.cs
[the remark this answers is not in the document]
Kaneraz: This is an open source project, not something we're just using internally. We should build toward what's possible in SQL Server, not just how we do it.

src/BackupChain.cs
Kaneraz: Exactly, outside the Facade we should not have references to SMO.

tests/AgDatabaseMove.Unit/FileToolsTest.cs
Kaneraz: Sorry, valid in SQL Server. I understand they're valid in this implementation.

src/SmoFacade/Server.cs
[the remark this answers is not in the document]
alexirion10: I need to debug & test to determine

src/BackupMetadata.cs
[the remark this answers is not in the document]
alexirion10: The project being open source means that anyone else is able to contribute and change these values to suit their needs.
Other extensions can always be added on.  I really don't think this is harmful checking a handful of extensions, but sure we can allow any file ext

src/BackupChain.cs
alexirion10: We need to use this type somewhere... Do you have a suggestion on where to use these types inside of the Facade?  Another wrapper class for an Enum type?
Would fully qualifying the Enum instead of `using` satisfy these concerns?

src/SmoFacade/BackupFileTools.cs
```diff
@@ -0,0 +1,49 @@
...
+    public enum BackupType
+    {
+      Full, // bak  / D
+      Diff, // diff / I
+      Log // trn  / L
+    }
+
+    public static bool IsUrl(string path)
+    {
+      return Regex.IsMatch(path, @"(http(|s):\/)(\/[a-z0-9\.\-]+)+\.([a-zA-Z]+)$");
```
Martin11175, line 18: Are we not using `try { new Uri(path) } catch { ... }` here to avoid non-HTTP URLs or something?
I feel like maintaining a regex for this is more complicated than it needs to be.

tests/AgDatabaseMove.Unit/FileToolsTest.cs
```diff
@@ -0,0 +1,60 @@
+namespace AgDatabaseMove.Unit
+{
+  using System.Collections.Generic;
+  using SmoFacade;
+  using Xunit;
+
+
+  public class BackupFileToolsTest
```
Martin11175, line 8: Do we want to add a couple of example failure cases here to as well to ensure full coverage?

Martin11175 (approved):
Lgtm and some healthy refactors along the way! 🏖️ ✔️

tests/AgDatabaseMove.Unit/FileToolsTest.cs
alexirion10: I have failure cases, check down below :)

src/SmoFacade/BackupFileTools.cs
alexirion10: Good idea, but I don't think `new Uri()` is going to throw for many of the cases that we want to block.
https://docs.microsoft.com/en-us/dotnet/api/system.uri?view=netcore-3.1
[This documentation](https://docs.microsoft.com/en-us/dotnet/api/system.uribuilder.scheme?view=netcore-3.1#remarks) says it, in fact, won't throw, but we could use the schema property to determine the type

PR #41, opened 2020-09-02
fix(regex): fix regex and cleanup after some testing

alexirion10:
Fixed regex to allow underscores, basically by defining a directory path as anything non-whitespace

Two TODO questions:

1. Should we consider allowing a SQL query for _backupPathTemplate?  Angelo's backups deployement scripts don't update `SMO.Server.BackupDirectory`.  Options to fix would be (1) update that registry value in Angelo's scripts (2) allow CLI option to pass in query for value (3) don't worry about it because the backupPathTemplate should be used
2. Should the source DB be altered to single-user mode prior to the move?  Is that expected to be done by the consuming code?

### DIFF
--- a/src/AgDatabase.cs
+++ b/src/AgDatabase.cs
@@ -25,7 +25,7 @@ namespace AgDatabaseMove
     List<BackupMetadata> RecentBackups();
     void JoinAg();
 
-    void Restore(IList<BackupMetadata> backupOrder, Func<string, string> fileRelocation = null);
+    void Restore(IEnumerable<BackupMetadata> backupOrder, Func<string, string> fileRelocation = null);
 
     void CopyLogins(IEnumerable<LoginProperties> logins);
     IEnumerable<LoginProperties> AssociatedLogins();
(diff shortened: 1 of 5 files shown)


src/AgDatabase.cs
```diff
@@ -25,7 +25,7 @@ public interface IAgDatabase
     List<BackupMetadata> RecentBackups();
     void JoinAg();
 
-    void Restore(IEnumerable<BackupMetadata> backupOrder, Func<string, string> fileRelocation = null);
+    void Restore(IList<BackupMetadata> backupOrder, Func<string, string> fileRelocation = null);
```
Kaneraz, line 28: I don't see any reason for this change. We don't need to add any items to the list as a part of the restore so IEnumerable will suffice and more accurately portray the intentions.

Kaneraz (changes requested):
The regex fix looks good. There's a lot fewer style/format changes to draw away from the purpose of the review, but again keeping them in a separate review is preferred. At least as a separate commit so they can be viewed with different levels of scrutiny.

We can talk more about the query option for a backup path template, that sounds interesting. I suspected that the template would just be a parameter on the CLI or defined in the consuming code. 

For single user mode, I envision the consumers doing that in their own code: https://github.com/factset/AgDatabaseMove/blob/master/tests/AgDatabaseMove.Integration/TestRestore.cs#L127

Kaneraz (approved):
Thanks, looks good!

PR #43, opened 2020-09-04
Feat(options): backup path sql query

alexirion10:
Instead of passing in a string of the backup path template, pass in a string that is a SQL query to fetch the location.
If no text is present, then the fallback is `_server.BackupDirectory`

The query allows more flexibility on where consumers store their backup paths.

Note: the default value for `BackupPathSqlQuery` could probably be removed as that's more CRDS specific

### DIFF
--- a/src/AgDatabase.cs
+++ b/src/AgDatabase.cs
@@ -51,16 +51,14 @@ namespace AgDatabaseMove
       Name = dbConfig.DatabaseName;
       _listener = new Listener(new SqlConnectionStringBuilder(dbConfig.ConnectionString) { InitialCatalog = "master" });
 
-      if(!string.IsNullOrWhiteSpace(dbConfig.BackupPathSqlQuery)) {
-        using (var cmd = _listener.Primary.SqlConnection.CreateCommand())
-        {
+      if(!string.IsNullOrWhiteSpace(dbConfig.BackupPathSqlQuery))
+        using(var cmd = _listener.Primary.SqlConnection.CreateCommand()) {
           cmd.CommandText = dbConfig.BackupPathSqlQuery;
-          using (var reader = cmd.ExecuteReader()) {
+          using(var reader = cmd.ExecuteReader()) {
             if(reader.Read())
               _backupPathTemplate = (string)reader[0];
           }
         }
-      }
     }
 
     /// <summary>
--- a/src/AgDatabase.cs
+++ b/src/AgDatabase.cs
@@ -39,7 +39,7 @@ namespace AgDatabaseMove
   /// </summary>
   public class AgDatabase : IDisposable, IAgDatabase
   {
-    private readonly string _backupPathTemplate;
+    private readonly string _backupPathSqlQuery;
     internal readonly IListener _listener;
 
     /// <summary>
@@ -49,16 +49,8 @@ namespace AgDatabaseMove
     public AgDatabase(DatabaseConfig dbConfig)
     {
       Name = dbConfig.DatabaseName;
+      _backupPathSqlQuery = dbConfig.BackupPathSqlQuery;
       _listener = new Listener(new SqlConnectionStringBuilder(dbConfig.ConnectionString) { InitialCatalog = "master" });
-
-      if(!string.IsNullOrWhiteSpace(dbConfig.BackupPathSqlQuery))
-        using(var cmd = _listener.Primary.SqlConnection.CreateCommand()) {
-          cmd.CommandText = dbConfig.BackupPathSqlQuery;
-          using(var reader = cmd.ExecuteReader()) {
-            if(reader.Read())
-              _backupPathTemplate = (string)reader[0];
-          }
-        }
     }
 
     /// <summary>
@@ -96,7 +88,7 @@ namespace AgDatabaseMove
     /// </summary>
     public void LogBackup()
     {
-      _listener.Primary.LogBackup(Name, _backupPathTemplate);
+      _listener.Primary.LogBackup(Name, _backupPathSqlQuery);
     }
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


src/SmoFacade/Server.cs
```diff
@@ -136,61 +136,51 @@ public Database Database(string dbName)
...
     {
-      var filePath = string.Format(backupPathTemplate,
-                                   backup.Database,
-                                   DateTime.Now.ToString("yyyy_MM_dd_hhmmss_fff"));
+      backupDirectoryPath = backupDirectoryPath ?? _server.BackupDirectory;
+      if(backupDirectoryPath.EndsWith("\\") || backupDirectoryPath.EndsWith("/"))
+        backupDirectoryPath = backupDirectoryPath.Substring(0, backupDirectoryPath.Length - 1);
+
+      var filePath =
+        $"{backupDirectoryPath}/{databaseName}/{databaseName}_backup_{DateTime.Now.ToString("yyyy_MM_dd_hhmmss_fff")}.{BackupFileTools.BackupTypeToExtension(type)}";
```
Kaneraz, line 176: The template allowed for the databaseName in the directory path as well as the file name, but this new implementation forces the directory per database setup and our naming conventions on all consumers.

src/AgDatabase.cs
```diff
@@ -50,7 +50,15 @@ public AgDatabase(DatabaseConfig dbConfig)
...
-      _backupPathTemplate = dbConfig.BackupPathTemplate;
+
+      if(!string.IsNullOrWhiteSpace(dbConfig.BackupPathSqlQuery))
+        using(var cmd = _listener.Primary.SqlConnection.CreateCommand()) {
+          cmd.CommandText = dbConfig.BackupPathSqlQuery;
+          using(var reader = cmd.ExecuteReader()) {
+            if(reader.Read())
+              _backupPathTemplate = (string)reader[0];
+          }
+        }
```
Kaneraz, line 61: Does this resolution belong at the AgDatabase level or should we do this at the server level since it could be different for each host? I lean towards leaving the resolution to the point where we take the backup as it prepares us for better handling of fail over scenarios and retries.

src/DatabaseConfig.cs
```diff
@@ -19,9 +19,11 @@ public class DatabaseConfig
...
     /// <summary>
-    ///   If this isn't supplied we build one from the database default backup path.
-    ///   See <see cref="Server.DefaultBackupPathTemplate" />
+    ///   SQL query text to retrieve the desired backup path location. If none is provided, Server.DefaultBackupPathTemplate is
+    ///   used.   See <see cref="Server.DefaultBackupPathTemplate" />
+    ///   <example> SELECT '\\my\backup\location\'</example>
+    ///   <example> SELECT Backup_path FROM [msdb].[dbo].[_Sys_Backup_config]</example>
     /// </summary>
-    public string BackupPathTemplate { get; set; }
+    public string BackupPathSqlQuery { get; set; } = "SELECT Backup_path FROM [msdb].[dbo].[_Sys_Backup_config]";
```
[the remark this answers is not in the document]
alexirion10: Right. The "note" in my PR comment above says this should probably be removed. I left it in as an example so that we understand what I'm trying to do for CRDS here.
The default, empty string, would be to execute the query you typed above. So you can leave the string empty instead of setting that string as the default

src/SmoFacade/Server.cs
alexirion10: From an interface perspective, there is no way to know what kind of template is required.  The only way to know how many template fields and what they are used for is to look at the source code.
I can remove the `"database name as a directory"`, but I'd really prefer to not have a templated string interface

src/AgDatabase.cs
alexirion10: Yeah, I can push the implementation down to `Server`.
re - "it could be different for each host": How do you want to handle that?  I was only going to use the Primary, which is what is being done [here](https://github.com/factset/AgDatabaseMove/blob/master/src/AgDatabase.cs#L91)

src/SmoFacade/Server.cs
Kaneraz: Ya, the templated string interface is pretty poor. It's just so easy.

Now that we're moving the directory path lookup from C# to SQL, should we just have the query generate the full backup and file path? I feel like even making the file name be databasename_backup_timestamp.bak is pretty prescriptive. If we're running within the context the database we're managing everything needed should be there.

src/AgDatabase.cs
Kaneraz: I don't think our current implementation handles a primary change during the AgDatabase object lifetime. Concerns about it working today on multiple hosts isn't at issue. It's just _if_ we made changes to support the [primary changing with fail over](https://github.com/factset/AgDatabaseMove/blob/master/src/SmoFacade/Listener.cs#L83) this means less state to manage or refactor out.

src/SmoFacade/Server.cs
alexirion10: I didn't do that because `_server.BackupDirectory` isn't going to return a templated date/filename.... although it would be nice to have control over the filename as well... hmm

src/AgDatabase.cs
alexirion10: I made these changes in the latest commit

alexirion10:
@Kaneraz for re-review

Martin11175 (approved):
Looks good to me!
Just a couple notes in case you think they're relevant.

src/SmoFacade/Server.cs
```diff
@@ -82,6 +82,31 @@ private DefaultFileLocations DefaultFileLocations()
...
+          cmd.CommandText = backupDirectoryQuery;
+          using(var reader = cmd.ExecuteReader()) {
+            if(reader.Read())
+              result = (string)reader[0];
+          }
+        }
+
+      result = result ?? _server.BackupDirectory;
+      if(result.EndsWith("\\") || result.EndsWith("/"))
+        result = result.Substring(0, result.Length - 1);
```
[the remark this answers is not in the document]
alexirion10: @Martin11175 - can you provide a snippet or example or docs?  Does the Path class consolidate double directory slashes?

Kaneraz (approved):

PR #44, opened 2020-09-11
feat(lang): c#8 and using statements

alexirion10:


Martin11175 (approved):
Neat! If it's using C#8 language features, what's the new minimum .NET Core / MSBuild Version supported?
Do we need to update any associated docs / build configs?

alexirion10:
@Martin11175 
1) https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version
2) uhhhh not that I know of.  I'm not going to update everything to C# lang version 8 immediately, will do so piece by piece as it makes sense.

PR #47, opened 2020-11-06
feat(restore): setting restore timeout command to 4 hours instead of 10 min default

alexirion10:
…default value of 10 minutes

### DIFF
--- a/src/SmoFacade/Server.cs
+++ b/src/SmoFacade/Server.cs
@@ -36,6 +36,8 @@ namespace AgDatabaseMove.SmoFacade
     public IEnumerable<Database> Databases => _server.Databases.Cast<Microsoft.SqlServer.Management.Smo.Database>()
       .Select(d => new Database(d, this));
 
+
+
     private AvailabilityGroup AvailabilityGroup =>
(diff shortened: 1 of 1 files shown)


src/SmoFacade/Server.cs
```diff
@@ -122,7 +122,7 @@ public Database Database(string dbName)
     public void Restore(IEnumerable<BackupMetadata> backupOrder, string databaseName,
       Func<string, string> fileRelocation = null)
     {
-      var restore = new Restore { Database = databaseName, NoRecovery = true };
+      var restore = new Restore { Database = databaseName, NoRecovery = true, RetryFailedQueries = true };
```
Kaneraz, line 125: Why set retry failed queries here? Are you just being explicit and it was already set by default?

Kaneraz (approved):

src/SmoFacade/Server.cs
alexirion10: boolean default is false. I thought this would be helpful, perhaps not?

Kaneraz: SMO default is true for this property. 
Helpful would depend on how the retry happens. If it'd mean a 4 hour timeout could try again and fail the job after 8 hours. May or may not be good.

PR #49, opened 2020-11-18
fix(drop): polly retry execute loop

alexirion10:
SMO & SqlClient upgrade, 
KillDatabase wrapped in retry loop

this is after more indepth testing with Angelo

### DIFF
--- a/src/SmoFacade/Database.cs
+++ b/src/SmoFacade/Database.cs
@@ -43,7 +43,7 @@ namespace AgDatabaseMove.SmoFacade
       var policy = Policy
         .Handle<FailedOperationException>()
         .Or<TimeoutException>()
-        .WaitAndRetry(6, retryAttempt => TimeSpan.FromMilliseconds(500 * retryAttempt));
+        .WaitAndRetry(6, retryAttempt => TimeSpan.FromMilliseconds(Math.Pow(10, retryAttempt)));
 
       // ensure database is not in AvailabilityGroup, WaitAndRetry loop for each instance to sync
       policy.Execute(() => {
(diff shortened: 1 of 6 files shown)


src/SmoFacade/Database.cs
```diff
@@ -95,6 +97,12 @@ public List<BackupMetadata> RecentBackups()
       return backups;
     }
 
+    public void SingleUserMode()
```
Kaneraz, line 100: Good addition, but not used?

Kaneraz (approved):
Looks like more aggressive retries, is the hope here to have be able to catch it on the retry before something else connects?

I know we made some timeout changes to support the longer running backups, do we need to make a similar change to drop so we can handle rollbacks of large (10min+) transactions?

src/SmoFacade/Database.cs
alexirion10: correct

alexirion10:
@Kaneraz - I was thinking over some of the math on the retries. Yes more aggressive, but since this is exponential the top end is very reasonable:
before (ms):
1000^1, 1000^2, 1000^3, 1000^4, 1000^5, 1000^6,
1000, 10000, 100000, 1000000, ...., .... ms
1, 10, 1000 (16.6667min), 10000 (166.66 min), 100000 (27.7 hours) seconds

after (ms):
10^1, 10^2, 10^3, 10^4, 10^5, 10^6
10, 100, 1000, 10000, 100000, 1000000 ms
.01, .1, 1, 10, 100 (1.66 min) , 1000 (16.6667 min) seconds

alexirion10:
also I added the retry logic to `Drop` in this PR. 
see `policy.Execute(() => { _database.Parent.KillDatabase(_database.Name); });`

Kaneraz:
Yep, 27 hours is a bit absurd

PR #50, opened 2020-11-23
feat(cred): explicit credential name

alexirion10:
SqlServer <= 2012 doesn't support SAS tokens for backup/restore operations. 
Credential name is required in this case
https://docs.microsoft.com/en-us/sql/relational-databases/backup-restore/sql-server-backup-to-url?view=sql-server-ver15#sqlserver


Kaneraz (approved):
LGTM, I presume the credentialName will be the same across all instances in the AG

alexirion10:
Yeah it'd have to be.  I haven't heard of any case where it wouldn't be, so long as all of the hosts are in the same Azure env (dev/qa/prod)

Issue #21, opened 2019-03-21
Connect to replicas with FQDN


Kaneraz:
When we connect to replicas with SSL we'll need to use the fully qualified domain name to match the certificate, but the replica server name we get from the AG is just the server name. We should be able to use `System.Net.Dns.GetHostEntry("hostname").HostName` to get the fqdn.

Kaneraz:
This should be resolved in #23

alexirion10:
This does not work with named instances

PR #51, opened 2020-11-23
fix(fqdn): hostname named instance #21

alexirion10:
DNS / fqdn name lookup doesn't work on Named Instances

#21

### DIFF
--- a/src/SmoFacade/Listener.cs
+++ b/src/SmoFacade/Listener.cs
@@ -75,7 +75,8 @@ namespace AgDatabaseMove.SmoFacade
       _secondaries = new List<Server>();
       foreach(var secondaryName in secondaryNames)
         _secondaries.Add(AgListenerNameToServer(ref connectionStringBuilder,
-                                                secondaryName, credentialName));
+                                                secondaryName,
+                                                credentialName));
     }
 
     public IEnumerable<Server> ReplicaInstances => Secondaries.Union(new[] { Primary });
@@ -129,7 +130,8 @@ namespace AgDatabaseMove.SmoFacade
       return dotIndex >= 0 ? dataSource.Remove(dotIndex) : dataSource;
     }
 
-    private static Server AgListenerNameToServer(ref SqlConnectionStringBuilder connBuilder, string agInstanceName, string credentialName)
+    private static Server AgListenerNameToServer(ref SqlConnectionStringBuilder connBuilder, string agInstanceName,
+      string credentialName)
     {
       var parts = agInstanceName.Split('\\');
       if(parts.Length == 1)
--- a/src/SmoFacade/Listener.cs
+++ b/src/SmoFacade/Listener.cs
@@ -69,12 +69,12 @@ namespace AgDatabaseMove.SmoFacade
       var secondaryNames = availabilityGroup.Replicas.Where(l => l != primaryName);
 
       // Connect to each server instance
-      Primary = AgListenerNameToServer(ref connectionStringBuilder, primaryName);
+      Primary = new Server(primaryName);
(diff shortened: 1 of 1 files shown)


Kaneraz (commented):
Have you tested this with an SSL connection per the original issue in #21?

alexirion10:
I have not.  Do you have any guides for that?
`Dns.GetHostEntry` throws an exception on named instances

Kaneraz:
It's in the connection string, I think encrypt=yes, but we should have it set in the consuming code.

src/SmoFacade/Listener.cs
```diff
@@ -128,11 +129,21 @@ private static string AgListenerName(string dataSource)
...
-    private static Server AgListenerNameToServer(ref SqlConnectionStringBuilder connBuilder, string agInstanceName,
-      string credentailName)
+    private static Server AgListenerNameToServer(ref SqlConnectionStringBuilder connBuilder, string agInstanceName)
     {
-      connBuilder.DataSource = Dns.GetHostEntry(agInstanceName).HostName;
-      return new Server(connBuilder.ToString(), credentailName);
+      try {
+        connBuilder.DataSource = Dns.GetHostEntry(agInstanceName).HostName;
+      }
+      catch(Exception) {
```
Kaneraz, line 137: Is there a more specific exception we should catch here?

[Probably irrelevant, see below]

src/SmoFacade/Listener.cs
```diff
@@ -128,11 +129,21 @@ private static string AgListenerName(string dataSource)
...
+        // NamedInstances: chop instance name, resolve DNS, slap instance name back on!
+        var parts = agInstanceName.Split('\\');
```
Kaneraz, line 139: We should just do this from the get go to determine if it's a named instance. 

What are the limitations on SQL hostnames and instance names?
https://docs.microsoft.com/en-us/previous-versions/sql/sql-server-2008-r2/ms143531(v=sql.105)?redirectedfrom=MSDN
https://docs.microsoft.com/en-us/troubleshoot/windows-server/identity/naming-conventions-for-computer-domain-site-ou#:~:text=The%20maximum%20length%20of%20the,from%20the%20NETBIOS%20host%20name.

They can't have a backslash in the name, so we should be good start with this.

Kaneraz (changes requested):
Looks like this resolves the potential issues with certificate errors. We can probably clean this up a bit more to avoid using exceptions for control flow.

src/SmoFacade/Listener.cs
alexirion10: will do.
I figured FDS names wouldn't have a `\`, but its reassuring that msft restricts it for all named instances

src/SmoFacade/Listener.cs
alexirion10: SocketException
https://docs.microsoft.com/en-us/dotnet/api/system.net.dns.gethostname?view=netcore-3.1

src/SmoFacade/Listener.cs
alexirion10: @Kaneraz - The problem with this actually is that `\` is not restricted from a servername, only a Named instance.
If I `Split` on backslash and it's a servername, what would happen?  

The exception control flow works nicely to only split the named instance if DNS doesn't match the first attempt

Kaneraz: I read it the other way, that our servername can't have a `\` in it.

alexirion10: gotcha, I guess neither can have `\` in the name

Kaneraz (approved):
Looks great, thanks!

PR #54, opened 2021-01-07
feat(login): drop login from all servers in Ag

alexirion10:


Martin11175:
Do we need to add this to the `AgDatabase.Delete()` function to actually request dropping the logins?

alexirion10:
We could... the way I wrote the code was like the following, decoupling the two from each other:
```C#
var agDb = new AgDatabase();
foreach(var loginProp in agDb.AssociatedLogins())
  agDb.DropLogin(loginProp);
agDb.Delete();
```

Martin11175:
That's fair, in which case:

- Should the `DropLogin()` method be added to the `IAgDatabase` to allow people to mock it out?
- Do we want a `DropAllLogins()` utility method as well?
- Could this method drop the login we're trying to use to do the move (and will that prevent us doing the `Delete()` call)?
- Should we add it as an option to the CLI for people not interacting programmatically?

alexirion10:
- Sure I can add it to the interface
- `DropAllLogins` is a possibility, but it's very easy for a consumer to do that themselves, I'm not sure it's entirely necessary
- Sooooo we're using this library not just for client moves anymore... its purpose has been expanded to any operations needed on an AG database. The case here is for a WipeDb operation
- I'm not adding it the CLI, as this functionality has nothing to do with database moves

Martin11175 (approved):
Fair enough! Questions answered and looks reasonable to me; :shipit:!

PR #55, opened 2021-01-13
chore(nuget): smo package and lower retries

alexirion10:
null handling
SMO package back down to 140
retry limit down to 4. 10^6ms is too long

### DIFF
--- a/src/AgDatabase.cs
+++ b/src/AgDatabase.cs
@@ -139,7 +139,7 @@ namespace AgDatabaseMove
 
     public IEnumerable<LoginProperties> AssociatedLogins()
     {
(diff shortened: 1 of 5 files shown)


Martin11175 (commented):
SMO updates and reducing retry timeouts to 10 seconds makes sense, not so sure on the `null` DB handling though; doesn't that mean that it'll silently return a `null` set from `AssociatedLogins` (not even an empty `IEnumerable`) and just not perform the `RestrictedUserMode()` action?  
I feel like if the DB doesn't exist that we're asking to perform the action on we should raise an exception of some kind (not just `NullReferenceException` though)?

PR #56, opened 2021-01-18
feat(database): Public constructor for the `Database` class

AdityaNPL:
So that we can create the `Database` object that is needed to initialise the `BackupChain` object

### DIFF
--- a/src/SmoFacade/Database.cs
+++ b/src/SmoFacade/Database.cs
@@ -1,118 +1,132 @@
 namespace AgDatabaseMove.SmoFacade
 {
-  using System;
-  using System.Collections.Generic;
-  using System.Linq;
-  using Microsoft.SqlServer.Management.Smo;
-  using Polly;
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using Microsoft.SqlServer.Management.Smo;
+    using Polly;
 
-
-  /// <summary>
-  ///   Adds some better accessors and simplifies some interactions with SMO's database class.
-  /// </summary>
-  public class Database
-  {
-    private readonly Microsoft.SqlServer.Management.Smo.Database _database;
-    private readonly Server _server;
-
-    public Database(Microsoft.SqlServer.Management.Smo.Database database, Server server)
-    {
-      _database = database;
-      _server = server;
-    }
-
-    public IEnumerable<User> Users => _database.Users.Cast<Microsoft.SqlServer.Management.Smo.User>()
-      .Select(u => new User(u, _server));
-
-    public string Name => _database.Name;
-
-    public bool Restoring => _database.Status == DatabaseStatus.Restoring;
-
-    public void RestoreWithRecovery()
-    {
-      // TODO: Sql injection paranoia. - can we execute normally with a parameterized statement here?
-      _database.Parent.Databases["master"].ExecuteNonQuery($"RESTORE DATABASE [{_database.Name}] WITH RECOVERY");
-      _database.Refresh();
-    }
 
     /// <summary>
-    ///   Kills connections and deletes the database.
+    ///   Adds some better accessors and simplifies some interactions with SMO's database class.
     /// </summary>
-    public void Drop()
+    public class Database
     {
-      var policy = Policy
-        .Handle<FailedOperationException>()
-        .Or<TimeoutException>()
-        .WaitAndRetry(4, retryAttempt => TimeSpan.FromMilliseconds(Math.Pow(10, retryAttempt)));
+        private readonly Microsoft.SqlServer.Management.Smo.Database _database;
+        private readonly Server _server;
 
-      // ensure database is not in AvailabilityGroup, WaitAndRetry loop for each instance to sync
-      policy.Execute(() => {
-        if(string.IsNullOrEmpty(_database.AvailabilityGroupName))
-          return;
-        if(_database.AvailabilityDatabaseSynchronizationState >
-           AvailabilityDatabaseSynchronizationState.NotSynchronizing)
-          throw new
-            TimeoutException($"Cannot kill the database {Name} until it has been removed from the AvailabilityGroup.");
-      });
+        public Database(DatabaseConfig databaseConfig) {
+            string connectionString = databaseConfig.ConnectionString;
+            string dbName = databaseConfig.DatabaseName;
 
-      policy.Execute(() => { _database.Parent.KillDatabase(_database.Name); });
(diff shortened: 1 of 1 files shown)


Martin11175 (commented):
What's the intended call pattern for this constructor?  
It'll work as-is but looks like you need one kind of database to construct the other kind of database; could we make the `public` constructor wrap the construction of the SMO DB object if the required parameters are simple enough?

AdityaNPL:
Yup there is a list of objects required to be created first before finally calling this constructor. 
For example, right now it will be something like this: 
```csharp
var serverConnection = new Microsoft.SqlServer.Management.Common.ServerConnection(host, username, password);
var microsoftServer = new Microsoft.SqlServer.Management.Smo.Server(serverConnection);
var microsoftDB = new Microsoft.SqlServer.Management.Smo.Database(microsoftServer, dbName);
var server = new Server(conString);
var db = new Database(microsoftDB, server);
```

Will try to see if we can have all of this within the public constructor instead. (but we might want to address it in another PR?)

AdityaNPL:
There are some problems with commits `e5385a1` and `3849a34` - looking to revert and reset them
Pls ignore for now

AdityaNPL:
forced push to reset and fix

src/SmoFacade/Database.cs
```diff
@@ -80,6 +93,7 @@ public List<BackupMetadata> RecentBackups()
       dbName.Value = _database.Name;
       cmd.Parameters.Add(dbName);
 
+      cmd.Connection.Open();
```
AdityaNPL, line 96: I had issues here as it said "SQLConnection was not open before executing a query".
Is this because it should already be opened on line 23?

AdityaNPL: Now getting an error when creating via the `AgDatabase` class - saying "SQLConnection is already open"

Wondering if we should open the connection for the `Database` class in it's constructor?

AdityaNPL: recent commit `591f14d` solves this issue - but want to know your thoughts on it.

src/SmoFacade/Database.cs
```diff
@@ -15,10 +15,35 @@ public class Database
...
     {
       _database = database;
       _server = server;
+      RefreshAndOpenSqlConnection();
+    }
+
+    ~Database()
+    {
+      _server.SqlConnection.Close();
+    }
```
alexirion10, line 42: `Dispose` is preferred to descriptor/finalizer in C#, which is garbage collected unlike C++:
https://stackoverflow.com/questions/4898733/when-should-i-create-a-destructor

src/SmoFacade/Database.cs
```diff
@@ -15,10 +15,35 @@ public class Database
...
+    private void RefreshAndOpenSqlConnection()
+    {
+      _server.SqlConnection.Close();
```
alexirion10, line 45: newline between methods, please.
I see that calling `Open` on an already open connection throws, so I suppose this is fine

src/SmoFacade/Database.cs
```diff
@@ -15,10 +15,35 @@ public class Database
...
 
+    public Database(DatabaseConfig databaseConfig)
+    { 
+        string connectionString = databaseConfig.ConnectionString;
+        string dbName = databaseConfig.DatabaseName;
+
+        var sqlConnection = new System.Data.SqlClient.SqlConnection(connectionString);
+        var sqlServerConnection = new Microsoft.SqlServer.Management.Common.ServerConnection(sqlConnection);
+        var sqlSmoServer = new Microsoft.SqlServer.Management.Smo.Server(sqlServerConnection);
+        _database = new Microsoft.SqlServer.Management.Smo.Database(sqlSmoServer, dbName);
```
alexirion10, line 26: put the following `using` statements at the top of this file so that these names aren't so verbose
```C#
  using System.Data.SqlClient;
  using Microsoft.SqlServer.Management.Common;
  using Microsoft.SqlServer.Management.Smo;
```

src/SmoFacade/Database.cs
```diff
@@ -15,10 +15,35 @@ public class Database
...
+
+        _server = new Server(connectionString);
+        RefreshAndOpenSqlConnection();
+    }
+
     internal Database(Microsoft.SqlServer.Management.Smo.Database database, Server server)
     {
       _database = database;
       _server = server;
+      RefreshAndOpenSqlConnection();
```
Kaneraz, line 36: Why do we need to add this to the original constructor?

src/SmoFacade/Database.cs
Kaneraz: Using the [iDisposable pattern](https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose) like Alex is suggesting is what's needed here if the database class needs to manage the sql connection lifetime. That's going to change how the object is used and managed by consumers as well as throughout the codebase.

Perhaps there's another way to allow consumers to manage the lifetime of the connection separate from the lifetime of the database object so we don't need to make this disposable though?

Kaneraz (commented):
If you make breaking changes like the database object being disposable make sure to increment the version appropriately and check our down stream dependencies for changes needed there as well.

src/SmoFacade/Database.cs
AdityaNPL: Oh wow didn't know that there was a big difference between c#/c++ destructor/finalizer
Had a look at some blogs and it looks like there are many caveats in how c# destructors are handled - which makes it very difficult to write a correct one.

Sure will transition to using the disposable pattern in c# from now on

AdityaNPL: I came up with an alternative implementation to fix the issue - it won't need this level of resource handling anymore, so looks to be a better solution.
What do you guys think?

(changes in most recent commit - `6aaf59c`)

src/SmoFacade/Database.cs
AdityaNPL: Not needed anymore

src/SmoFacade/Database.cs
AdityaNPL: ^^ scratch the above
I realised this is a long winded way of doing it - consumer can just do the below instead to achieve the same
`Database db = new Server(connString).Database(dbName)`

AdityaNPL:
Closing PR as this feature is not required anymore.

As mentioned in one of the comments above, consumers can use the `Database()` method in the `SmoFacade.Server` class to get the `SmoFacade.Database` object.
So can do something like:
```csharp
SmoFacade.Database db = new SmoFacade.Server(connectionString).Database(dbName);
```

PR #57, opened 2021-01-26
fix(linux_dns_issue): Re-trying DNS lookup of instance name after appending the domain from AG listener

AdityaNPL:
`Dns.GetHostEntry(instance)` fails intermittently on linux platforms if `instance` doesn't have the full domain

### DIFF
--- a/src/SmoFacade/Listener.cs
+++ b/src/SmoFacade/Listener.cs
@@ -5,7 +5,8 @@ namespace AgDatabaseMove.SmoFacade
   using System.Data.SqlClient;
   using System.Linq;
   using System.Net;
-  using System.Threading.Tasks;
+    using System.Runtime.InteropServices;
+    using System.Threading.Tasks;
 
 
   internal interface IListener : IDisposable
@@ -134,15 +135,27 @@ namespace AgDatabaseMove.SmoFacade
       string credentialName)
     {
       var parts = agInstanceName.Split('\\');
-      if(parts.Length == 1)
-        connBuilder.DataSource = Dns.GetHostEntry(agInstanceName).HostName;
-      else if(parts.Length == 2)
-        // NamedInstances: chop instance name, resolve DNS, slap instance name back on!
-        connBuilder.DataSource = $"{Dns.GetHostEntry(parts[0]).HostName}\\{parts[1]}";
-      else
-        throw new ArgumentException($"agInstanceName param {agInstanceName} cannot be resolved by DNS");
+            if (parts.Length == 1)
+                connBuilder.DataSource = ResolveDnsAndGetHostName(connBuilder.DataSource, agInstanceName);
+            else if (parts.Length == 2)
+                // NamedInstances: chop instance name, resolve DNS, slap instance name back on!
+                connBuilder.DataSource = $"{ResolveDnsAndGetHostName(connBuilder.DataSource, parts[0])}\\{parts[1]}";
+            else
+                throw new ArgumentException($"agInstanceName param {agInstanceName} cannot be resolved by DNS");
 
       return new Server(connBuilder.ToString(), credentialName);
+    }
+
+    private static string ResolveDnsAndGetHostName(string hostUrl, string instance)
+    {
+      if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux) && !instance.Contains('.'))
+      {
+        string[] urlFragments = hostUrl.Split('.');
+        urlFragments[0] = "";
+        instance += string.Join(".", urlFragments);
+      }
+      var res = Dns.GetHostEntry(instance).HostName;
+      return res;
     }
   }
 }
--- a/src/SmoFacade/Listener.cs
+++ b/src/SmoFacade/Listener.cs
@@ -5,8 +5,7 @@ namespace AgDatabaseMove.SmoFacade
   using System.Data.SqlClient;
   using System.Linq;
   using System.Net;
-    using System.Runtime.InteropServices;
-    using System.Threading.Tasks;
+  using System.Threading.Tasks;
 
 
   internal interface IListener : IDisposable
@@ -146,16 +145,23 @@ namespace AgDatabaseMove.SmoFacade
       return new Server(connBuilder.ToString(), credentialName);
     }
 
-    private static string ResolveDnsAndGetHostName(string hostUrl, string instance)
+    /// <summary>
+    ///   Resolves "instanceName" to a FQDNS name
+    ///   However on Unix OS, "instanceName" should be a complete domain - "{sub domain}.{second level domain}.{...}.{top level domain}" (eg: "xyz.abc.def.com")
+    ///   Therefore, append from the {second-level} -> {top-level} domains from the "hostDomain" to the "instanceName" to make it a complete domain if it is not already
+    /// </summary>
+    /// <param name="hostDomain">The complete domain for the SQL server</param>
+    /// <param name="instanceName">The specific name of an instance within the server</param>
+    private static string ResolveDnsAndGetHostName(string hostDomain, string instanceName)
     {
-      if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux) && !instance.Contains('.'))
+      if (Environment.OSVersion.Platform == PlatformID.Unix && !instanceName.Contains('.'))
       {
-        string[] urlFragments = hostUrl.Split('.');
-        urlFragments[0] = "";
-        instance += string.Join(".", urlFragments);
+        string[] domainFragments = hostDomain.Split('.');
+        domainFragments[0] = "";
+        instanceName += string.Join(".", domainFragments);
       }
-      var res = Dns.GetHostEntry(instance).HostName;
-      return res;
+      return Dns.GetHostEntry(instanceName).HostName;
+
     }
   }
 }
--- a/src/SmoFacade/Listener.cs
+++ b/src/SmoFacade/Listener.cs
@@ -133,16 +133,15 @@ namespace AgDatabaseMove.SmoFacade
     private static Server AgListenerNameToServer(ref SqlConnectionStringBuilder connBuilder, string agInstanceName,
       string credentialName)
     {
-      var parts = agInstanceName.Split('\\');
-      if(parts.Length == 1)
+      try 
+      {
         connBuilder.DataSource = ResolveDnsAndGetHostName(connBuilder.DataSource, agInstanceName);
-      else if(parts.Length == 2)
-        // NamedInstances: chop instance name, resolve DNS, slap instance name back on!
-        connBuilder.DataSource = $"{ResolveDnsAndGetHostName(connBuilder.DataSource, parts[0])}\\{parts[1]}";
-      else
-        throw new ArgumentException($"agInstanceName param {agInstanceName} cannot be resolved by DNS");
-
-      return new Server(connBuilder.ToString(), credentialName);
+        return new Server(connBuilder.ToString(), credentialName);
+      }
+      catch (Exception e)
+      {
+        throw new ArgumentException($"agInstanceName param {agInstanceName} cannot be resolved by DNS", e);
+      }
     }
 
     /// <summary>
@@ -154,33 +153,47 @@ namespace AgDatabaseMove.SmoFacade
     /// <param name="agReplicaInstanceName">The specific name of a replica instance within the AG</param>
     private static string ResolveDnsAndGetHostName(string agListenerDomain, string agReplicaInstanceName)
     {
-      Tuple<string, string> domain_port = SplitDomainPort(agListenerDomain);
-      if (Environment.OSVersion.Platform == PlatformID.Unix)
+      // sometimes instances/listners have ports and named instances. Therefore, we strip them off before calling DNS.GetHostEntry() and then add them back to the result 
+      (string listenerDomain, string listenerPort) = SplitDomainPort(agListenerDomain);
+      (string instanceDomain, string instancePort) = SplitDomainPort(agReplicaInstanceName);
+      // we want to add back instancePort if it exists, otherwise the server port
+      string portToUse = instancePort ?? listenerPort;
+      try
       {
-        string[] domainFragments = domain_port.Item1.Split('.');
-        domainFragments[0] = "";
-        agReplicaInstanceName += string.Join(".", domainFragments);
+        return $"{Dns.GetHostEntry(instanceDomain).HostName}{portToUse}";
       }
-      return Dns.GetHostEntry(agReplicaInstanceName).HostName+domain_port.Item2;
+      catch (System.Net.Sockets.SocketException e)
+      {
+        Console.WriteLine($"Unable to get dns for instance '{instanceDomain}'\n{e}");
 
+        // Re-try with full domain appended to the instanceDomain by stripping off domain from listner and appending to instance
+        // eg: if listner is "abc.def.ghi" we want to append ".def.ghi" to the instance
+        string[] domainFragments = listenerDomain.Split('.');
+        domainFragments[0] = "";
+        string instanceDomainFull = $"{instanceDomain}{string.Join(".", domainFragments)}";
+
+        Console.WriteLine($"Retrying with full domain '{instanceDomainFull}");
+        return $"{Dns.GetHostEntry(instanceDomainFull).HostName}{portToUse}";
+      }
     }
-    private static Tuple<string,string> SplitDomainPort(string domainAndPort)
+
+    private static (string domain, string port) SplitDomainPort(string domainAndPort)
     {
-      string domian = domainAndPort;
-      string port = "";
+      string domain = domainAndPort;
+      string port = null;
       if (domainAndPort.Contains(','))
       {
         string[] fragments = domainAndPort.Split(new char[] {','}, 2);
-        domian = fragments[0];
-        port = ","+fragments[1];
+        domain = fragments[0];
+        port = $",{fragments[1]}";
       }
-      else if (domainAndPort.Contains('\\'))
+      else if (domainAndPort.Contains('\\')) // handling named instances the same way as ports
       {
         string[] fragments = domainAndPort.Split(new char[] {'\\'}, 2);
-        domian = fragments[0];
-        port = "\\"+fragments[1];
+        domain = fragments[0];
+        port = $"\\{fragments[1]}";
       }
-      return new Tuple<string, string>(domian, port);
+      return (domain, port);
     }
   }
 }
--- a/src/SmoFacade/Listener.cs
+++ b/src/SmoFacade/Listener.cs
@@ -134,13 +134,13 @@ namespace AgDatabaseMove.SmoFacade
       string credentialName)
     {
       var parts = agInstanceName.Split('\\');
-        if(parts.Length == 1)
-          connBuilder.DataSource = ResolveDnsAndGetHostName(connBuilder.DataSource, agInstanceName);
-        else if(parts.Length == 2)
-          // NamedInstances: chop instance name, resolve DNS, slap instance name back on!
-          connBuilder.DataSource = $"{ResolveDnsAndGetHostName(connBuilder.DataSource, parts[0])}\\{parts[1]}";
-        else
-          throw new ArgumentException($"agInstanceName param {agInstanceName} cannot be resolved by DNS");
+      if(parts.Length == 1)
+        connBuilder.DataSource = ResolveDnsAndGetHostName(connBuilder.DataSource, agInstanceName);
+      else if(parts.Length == 2)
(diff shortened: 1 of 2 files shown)


src/SmoFacade/Listener.cs
```diff
@@ -134,15 +135,27 @@ private static string AgListenerName(string dataSource)
...
+            else if (parts.Length == 2)
+                // NamedInstances: chop instance name, resolve DNS, slap instance name back on!
+                connBuilder.DataSource = $"{ResolveDnsAndGetHostName(connBuilder.DataSource, parts[0])}\\{parts[1]}";
+            else
+                throw new ArgumentException($"agInstanceName param {agInstanceName} cannot be resolved by DNS");
 
       return new Server(connBuilder.ToString(), credentialName);
     }
+
+    private static string ResolveDnsAndGetHostName(string hostUrl, string instance)
```
Kaneraz, line 149: This looks like what @am99 was suggesting, but it's unclear here the difference between the instance and hostUrl. Let's rename and/or document the difference here.

Kaneraz (commented):
Some odd spacing showing up in the diff, but it looks like it's doing what we set out to do.

src/SmoFacade/Listener.cs
AdityaNPL: Sure have renamed and added some summary comments - let me know if you guys want the wording changed

src/SmoFacade/Listener.cs
```diff
@@ -135,14 +135,33 @@ private static string AgListenerName(string dataSource)
...
+    /// <summary>
+    ///   Resolves "instanceName" to a FQDNS name
+    ///   However on Unix OS, "instanceName" should be a complete domain - "{sub domain}.{second level domain}.{...}.{top level domain}" (eg: "xyz.abc.def.com")
+    ///   Therefore, append from the {second-level} -> {top-level} domains from the "hostDomain" to the "instanceName" to make it a complete domain if it is not already
+    /// </summary>
+    /// <param name="hostDomain">The complete domain for the SQL server</param>
+    /// <param name="instanceName">The specific name of an instance within the server</param>
+    private static string ResolveDnsAndGetHostName(string hostDomain, string instanceName)
+    {
+      if (Environment.OSVersion.Platform == PlatformID.Unix && !instanceName.Contains('.'))
```
AdityaNPL, line 157: had to change 
from `RuntimeInformation.IsOSPlatform(OSPlatform.Linux)` 
to `Environment.OSVersion.Platform == PlatformID.Unix`
due to .Net Framework 4.7 compatibility issues

alexirion10 (commented):
Code looks ok... but this is tricky. 
I'm not certain that `instanceName.Contains('.')` means that it's a FQDN necessarily
Did you test?  How did you test?  Can you provide some sample I/O?

AdityaNPL:
I'm not sure if there is another way to check? I thought that if it has at least one `.` it could already be a FQDN?
Do we want to ignore if it is or not and just blindly append the later part of the domain from the `hostDomain` to the instance name extracted from `instanceName`?

i.e. if `hostDomain = "abc.def.ghi"` and `instanceName="uvw.xyz"` we do dns lookup on `uvw.def.ghi` ??

alexirion10:
Exactly...
I don't have an answer to that. I'm asking if you've tried and thought through all of these possibilities

AdityaNPL:
@alexirion10 checked where this function is being called from and it looks like the only values being passed in are the replica instance names - which won't be FQDN
Have removed the check as I feel we don't need it anymore

Also renamed a few names as I am understanding the terminologies and tech stack a little better now 😄 

What do you think?

src/SmoFacade/Listener.cs
```diff
@@ -135,14 +135,52 @@ private static string AgListenerName(string dataSource)
...
+    /// </summary>
+    /// <param name="agListenerDomain">The complete domain for the SQL server AG listener</param>
+    /// <param name="agReplicaInstanceName">The specific name of a replica instance within the AG</param>
+    private static string ResolveDnsAndGetHostName(string agListenerDomain, string agReplicaInstanceName)
+    {
+      Tuple<string, string> domain_port = SplitDomainPort(agListenerDomain);
+      if (Environment.OSVersion.Platform == PlatformID.Unix)
+      {
+        string[] domainFragments = domain_port.Item1.Split('.');
+        domainFragments[0] = "";
```
alexirion10, line 161: Why are you setting domainFragments[0] to ""?  This seems random... would like a comment here or soemthing

src/SmoFacade/Listener.cs
```diff
@@ -135,14 +135,52 @@ private static string AgListenerName(string dataSource)
...
+    /// <summary>
+    ///   Resolves "agReplicaInstanceName" to a FQDNS name
+    ///   However on Unix OS, "agReplicaInstanceName" should be a complete domain - "{sub domain}.{second level domain}.{...}.{top level domain}" (eg: "xyz.abc.def.com")
+    ///   Therefore, we use {second-level} -> {top-level} part of the domain from "agListenerDomain" and append it to "agReplicaInstanceName"
+    /// </summary>
+    /// <param name="agListenerDomain">The complete domain for the SQL server AG listener</param>
+    /// <param name="agReplicaInstanceName">The specific name of a replica instance within the AG</param>
+    private static string ResolveDnsAndGetHostName(string agListenerDomain, string agReplicaInstanceName)
+    {
+      Tuple<string, string> domain_port = SplitDomainPort(agListenerDomain);
```
alexirion10, line 157: no snake_case please! we use camelCase everywhere else in the code

alexirion10 (changes requested):

Kaneraz:
@AdityaNPL if we're concerned about getting the FQDN instead of just a hostname we can attempt to resolve the hostname first and if it fails, then attempt to append from the original listener name.

AdityaNPL:
Changing this to Draft as I think this needs much more refactoring before it is ready to release publicly.
Leaving PR open to continue with reviews and suggestions.

Sure @Kaneraz , yes I think changing to a try-catch might be less error prone for us right now - will change to that

src/SmoFacade/Listener.cs
AdityaNPL: sure will put in a comment there

It's to strip off the first part of the domain before appending it to the instance name
so if it's like `abc.def.ghi` we want only `.def.ghi` to be appended to the instance name

I think we should end up using `UriBuilder` class in the end rather than our own string manipulations:
https://docs.microsoft.com/en-us/dotnet/api/system.uribuilder?view=netcore-2.0

src/SmoFacade/Listener.cs
```diff
@@ -135,14 +135,52 @@ private static string AgListenerName(string dataSource)
...
+      if (Environment.OSVersion.Platform == PlatformID.Unix)
+      {
+        string[] domainFragments = domain_port.Item1.Split('.');
+        domainFragments[0] = "";
+        agReplicaInstanceName += string.Join(".", domainFragments);
+      }
+      return Dns.GetHostEntry(agReplicaInstanceName).HostName+domain_port.Item2;
+
+    }
+    private static Tuple<string,string> SplitDomainPort(string domainAndPort)
```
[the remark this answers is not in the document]
AdityaNPL: looks like in c# I can actually create a tuple (like python) and I can access it via the names directly

It's called tuple types - what do you think about these?
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/value-tuples

src/SmoFacade/Listener.cs
```diff
@@ -133,16 +133,15 @@ private static string AgListenerName(string dataSource)
     private static Server AgListenerNameToServer(ref SqlConnectionStringBuilder connBuilder, string agInstanceName,
       string credentialName)
     {
-      var parts = agInstanceName.Split('\\');
-      if(parts.Length == 1)
```
AdityaNPL, line 137: `SplitDomainPort()` will now handle this

src/SmoFacade/Listener.cs
```diff
@@ -154,33 +153,47 @@ private static string AgListenerName(string dataSource)
...
+      {
+        Console.WriteLine($"Unable to get dns for instance '{instanceDomain}'\n{e}");
+
+        // Re-try with full domain appended to the instanceDomain by stripping off domain from listner and appending to instance
+        // eg: if listner is "abc.def.ghi" we want to append ".def.ghi" to the instance
+        string[] domainFragments = listenerDomain.Split('.');
+        domainFragments[0] = "";
+        string instanceDomainFull = $"{instanceDomain}{string.Join(".", domainFragments)}";
 
+        Console.WriteLine($"Retrying with full domain '{instanceDomainFull}");
```
AdityaNPL, line 175: Do we want these logging messages in the final release - I think they would be helpful for debugging for anyone using this

What do you guys think?

src/SmoFacade/Listener.cs
```diff
@@ -135,14 +135,52 @@ private static string AgListenerName(string dataSource)
...
+      {
+        string[] fragments = domainAndPort.Split(new char[] {','}, 2);
+        domian = fragments[0];
+        port = ","+fragments[1];
+      }
+      else if (domainAndPort.Contains('\\'))
+      {
+        string[] fragments = domainAndPort.Split(new char[] {'\\'}, 2);
+        domian = fragments[0];
+        port = "\\"+fragments[1];
```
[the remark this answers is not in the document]
AdityaNPL: done

src/SmoFacade/Listener.cs
```diff
@@ -133,16 +133,67 @@ private static string AgListenerName(string dataSource)
...
+        domain = fragments[0];
+        port = $",{fragments[1]}";
+      }
+      else if (domainAndPort.Contains('\\')) // handling named instances the same way as ports
+      {
+        string[] fragments = domainAndPort.Split(new char[] {'\\'}, 2);
+        domain = fragments[0];
+        port = $"\\{fragments[1]}";
+      }
+      return (domain, port);
```
alexirion10, line 196: This method needs unit tests

```C#
var splitChar = domainAndPort.Contains(',') ? ',' : domainAndPort.Contains('\\') ? ' \\' : null;
if(splitChar != null)
{
  string[] fragments = domainAndPort.Split(splitChar, 2);
  domain = fragments[0];
  port = $"{splitChar}{fragments[1]}";
}

return (domain, port);

```

src/SmoFacade/Listener.cs
alexirion10: Console logging isn't desireable in a Nuget package release IMO.
If you want log statements, they should go through Microsoft ILogger, for which there are many extensions to... like nlog
https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/?view=aspnetcore-5.0

src/SmoFacade/Listener.cs
AdityaNPL: ^^ ah yes, much better way of doing it - thanks! 😊 

Yes agreed, there can be many edge cases here - will look at creating some unit tests

Was just implementing the PR suggestions from above. 
For next step, as mentioned before, I'm going to look at using this class instead of doing the fragmentations myself: 
[`UriBuilder`](https://docs.microsoft.com/en-us/dotnet/api/system.uribuilder?view=netcore-2.0)

AdityaNPL: For `UriBuilder`, looks like it expects a well formatted URI (with protocol), which we don't have.
Plus I can't see functions there to segment out the domain name Itself.
(for ports, we are also handling named instances - which is SQL Sever specific so it won't be able to handle those either)

reverting back to using normal string split functions.

src/SmoFacade/Listener.cs
AdityaNPL: Had a look at `ILogger` and looks like it might be better to implement that (if we agree to do so) in it's own PR. There are many components that we will need to create for it to function (Logger models, categories, logging levels, etc.)

Shelfing for now, will come back to it if we feel it is necessary

src/SmoFacade/Listener.cs
```diff
@@ -130,19 +130,70 @@ private static string AgListenerName(string dataSource)
...
+    /// <param name="agReplicaInstanceName"> The name for an instance within the AG (for which we are trying to get the FQDN)</param>
+    /// <param name="agListenerDomain"> The complete domain for the AG listener</param>
+    private static string ResolveDnsHostNameForInstance(string agReplicaInstanceName, string agListenerDomain)
+    {
+      // Sometimes instances and listeners have ports(or "named instances")
+      // Therefore, we strip them off before calling DNS.GetHostEntry() and then add them back to the result 
+      var (listenerDomain, listenerPort) = SplitDomainAndPort(agListenerDomain);
+      var (instanceName, instancePort) = SplitDomainAndPort(agReplicaInstanceName);
+      // preference is to add back 'instancePort' over 'listenerPort' (in almost all cases they should be identical)
+      var port = instancePort ?? listenerPort;
```
AdityaNPL, line 161: Had a chat with Chris and we should look to change priority here - we want to prefer the port over named instance.
Mainly because ports are TCP/IP standards while named instances are only SQL Server standards. 

What do you guys think?

(should also rename `instancePort` to `instancePortOrNamedInstance` so that it is clear 😅 )

src/SmoFacade/Listener.cs
```diff
@@ -130,19 +130,70 @@ private static string AgListenerName(string dataSource)
...
+        listenerDomainFragments[0] = null;
+        var instanceDomain = $"{instanceName}{string.Join(".", listenerDomainFragments)}";
+
+        return $"{Dns.GetHostEntry(instanceDomain).HostName}{port}";
+      }
     }
+
+    // This function handles named instances ("<domain>\<named instance>") in the same way as ports
+    private static (string, string) SplitDomainAndPort(string domainAndPort)
+    {
```
AdityaNPL, line 182: @alexirion10 working on Unit tests for this - was wondering if we should have some for `ResolveDnsHostNameForInstance()` too to check it retries correctly?

Or is just checking this one fine?

alexirion10: the more tests the better

AdityaNPL:
Have refactored and added tests for the utility functions - I think should be ready for release
@alexirion10 / @Kaneraz can you pls give it a look over and let me know if you have any suggestions

src/SmoFacade/Listener.cs
```diff
@@ -130,19 +131,81 @@ private static string AgListenerName(string dataSource)
...
+    ///   Therefore, if dns lookup on just the instance name fails, we retry after appending the domain fragments from the listener to the instance name
+    /// </summary>
+    /// <param name="agReplicaInstanceName"> The name for an instance within the AG (for which we are trying to get the FQDN)</param>
+    /// <param name="agListenerDomain"> The complete domain for the AG listener</param>
+    private static string ResolveDnsHostNameForInstance(string agReplicaInstanceName, string agListenerDomain)
+    {
+      // Sometimes instances and listeners have ports or "named instances" (NI)
+      // Therefore, we strip them off before calling DNS.GetHostEntry() and then add them back to the result 
+      var (listenerDomain, listenerPortOrNi) = SplitDomainAndPort(agListenerDomain);
+      var (instanceName, instancePortOrNi) = SplitDomainAndPort(agReplicaInstanceName);
```
alexirion10, line 160: I'm not big on acronyms in variable names... `Ni` --> `NamedInstance` is verbose but more readable

tests/AgDatabaseMove.Unit/SmoFacadeListenerTest.cs
```diff
@@ -0,0 +1,86 @@
...
+
+    // {input, expectedDomain, expectedPort (can be null)}
+    public static IEnumerable<object[]> ValidPorts => new List<object[]> {
+      new object[] { $"{DOMAIN}", DOMAIN },
+      new object[] { $"{DOMAIN},123", DOMAIN, ",123" },
+      new object[] { $"{DOMAIN}\\SQL", DOMAIN, "\\SQL" },
+
+      new object[] { $"{HOST}", HOST },
+      new object[] { $"{HOST},123", HOST, ",123" },
+      new object[] { $"{HOST}\\SQL", HOST, "\\SQL" }
```
alexirion10, line 24: No need to interpolate $"{DOMAIN}" and $"{HOST} when it's already a string and not being concatenated.
Some other strings like ",123" and "\\SQL" could be put in const string variables too

tests/AgDatabaseMove.Unit/SmoFacadeListenerTest.cs
```diff
@@ -0,0 +1,86 @@
...
+    };
+
+    [Theory]
+    [MemberData(nameof(ValidPorts))]
```
alexirion10, line 28: `[InlineData]` is a bit more simple if you're only going to use the Enumerable inputs once
https://hamidmosalla.com/2017/02/25/xunit-theory-working-with-inlinedata-memberdata-classdata/

alexirion10 (commented):
I mean... I'm sure it works, but man this code is messy and fragile. I'm nervous with the amount of string parsing happening in here

src/SmoFacade/Listener.cs
AdityaNPL: I was debating this myself, it looked very long - so thought it might be better to name it `Ni` and put in a comment above
`// Sometimes instances and listeners have ports or "named instances" (NI)`

But yes, I can see your point and I now agree too - should go for more easy to read names

tests/AgDatabaseMove.Unit/SmoFacadeListenerTest.cs
AdityaNPL: Looks like we can't use `[InlineData]` for the first 2 tests - it doesn't support interpolated strings
Will change the last one

AdityaNPL:
The code logic of our fix:
1. Strip off any ports/named instances from both the listener and the instance
2. Try to resolve to FQDN by just using the instance name
3. If it fails (as it does sometimes on linux) we append the domain fragments from the listener to the instance name (to make it a complete domain) and then re-try
4. To the resolved FQDN, add back a port or a named instance (if available from step 1)

@alexirion10 I agree, but looking above at the steps we need to do, I think the string manipulations are required. 
I did have a look at using some in-built c# classes to perform these actions as well (like `UriBuilder` and `SqlConnectionStringBuilder`) but couldn't find any that could

---

I have some suggestions that could give us more confidence:
- I can intensify testing a bit more?
  - For example: refactor the string manipulation in `ResolveDnsHostNameForInstance()` into its own function and write some tests for that?
- We can also add some string format checks to make sure inputs are in correct format before we start splitting and slicing?
- We could go down the path of `regex` - but I opted not to because it is usually more complex to develop and felt it might introduce more bugs for something we can do with some simple string manipulations (which are quicker to implement and easier to follow) 

What are your thoughts?

Kaneraz (approved):
As we've discussed this is a bit of a tangle, but after looking at it a few time I'm not seeing a clear way to untangle it. Once we get things in place to resolve WINS on the containers this runs in, we can consider removing the hack.

Thanks for the unit tests, they give me confidence that this is going to remain stable as we keep moving forward and balances out my concerns with how tangled it feels.

src/SmoFacade/Listener.cs
```diff
@@ -133,16 +133,67 @@ private static string AgListenerName(string dataSource)
...
+        return new Server(connBuilder.ToString(), credentialName);
+      }
+      catch (Exception e)
+      {
+        throw new ArgumentException($"agInstanceName param {agInstanceName} cannot be resolved by DNS", e);
+      }
+    }
+
+    /// <summary>
+    ///   Resolves "agReplicaInstanceName" to a FQDNS name
```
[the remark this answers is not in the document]
AdityaNPL: looks like this is already fixed
(this is from an old commit)

src/SmoFacade/Listener.cs
```diff
@@ -1,3 +1,4 @@
+[assembly: System.Runtime.CompilerServices.InternalsVisibleTo("AgDatabaseMove.Unit")]
```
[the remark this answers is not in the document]
AdityaNPL: PR: https://github.com/factset/AgDatabaseMove/pull/63

src/SmoFacade/Listener.cs
```diff
@@ -133,16 +133,67 @@ private static string AgListenerName(string dataSource)
...
+      }
+      catch (Exception e)
+      {
+        throw new ArgumentException($"agInstanceName param {agInstanceName} cannot be resolved by DNS", e);
+      }
+    }
+
+    /// <summary>
+    ///   Resolves "agReplicaInstanceName" to a FQDNS name
+    ///   However on Unix OS, "agReplicaInstanceName" should be a complete domain - "{sub domain}.{second level domain}.{...}.{top level domain}" (eg: "xyz.abc.def.com")
```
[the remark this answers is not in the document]
AdityaNPL: PR: https://github.com/factset/AgDatabaseMove/pull/63

PR #59, opened 2021-02-16
fix(copy_only backups): ignore copy_only backups when making the backup-chain

AdityaNPL:
Edited SQL query to ignore `copy_only` backups when creating the backup chain

---

Tested against a local MSSQL instance.

Created a normal Full backup and then a `copy_only` Full backup
Chain started from the correct backup point and didn't contain any `copy_only` backups

### DIFF
--- a/src/SmoFacade/Database.cs
+++ b/src/SmoFacade/Database.cs
@@ -61,6 +61,11 @@ namespace AgDatabaseMove.SmoFacade
     /// <summary>
     ///   Queries msdb on the instance for backups of this database.
     /// </summary>
+    /// <param name="ignoreCopyOnlyBackups">
+    ///   Setting this to `true` will ignore copy_only backups
+    ///   These are temp backups made for a specific use case and should not be part of the backup-chain for general restore processes
+    ///   For more info: https://docs.microsoft.com/sql/relational-databases/backup-restore/copy-only-backups-sql-server
+    /// </param>
     /// <returns>A list of backups known by msdb</returns>
     public List<BackupMetadata> RecentBackups(bool ignoreCopyOnlyBackups=false)
     {
--- a/src/SmoFacade/Database.cs
+++ b/src/SmoFacade/Database.cs
@@ -61,13 +61,8 @@ namespace AgDatabaseMove.SmoFacade
     /// <summary>
     ///   Queries msdb on the instance for backups of this database.
     /// </summary>
-    /// <param name="ignoreCopyOnlyBackups">
-    ///   Setting this to `true` will ignore copy_only backups
-    ///   These are temp backups made for a specific use case and should not be part of the backup-chain for general restore processes
-    ///   For more info: https://docs.microsoft.com/sql/relational-databases/backup-restore/copy-only-backups-sql-server
-    /// </param>
     /// <returns>A list of backups known by msdb</returns>
-    public List<BackupMetadata> RecentBackups(bool ignoreCopyOnlyBackups=false)
+    public List<BackupMetadata> RecentBackups()
     {
       var backups = new List<BackupMetadata>();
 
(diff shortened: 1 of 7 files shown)


Martin11175 (approved):
Looks reasonable to me! Might it be worth adding a quick docstring for the `ignoreCopyOnlyBackups` parameter explaining why someone might want to set it (beyond just "it ignores `is_copy_only` backups")?

alexirion10 (changes requested):
I think a better approach here is the SQL query, not filtering out is_copy_only, but instead returning an extra column "is_copy_only".
In C# code when we can filter out those results when building the backup chain

src/SmoFacade/Database.cs
```diff
@@ -61,17 +61,29 @@ public void Drop()
...
                   "s.database_backup_lsn, s.checkpoint_lsn, s.[type] AS backup_type, s.server_name, s.recovery_model " +
                   "FROM msdb.dbo.backupset s " +
                   "INNER JOIN msdb.dbo.backupmediafamily m ON s.media_set_id = m.media_set_id " +
-                  "WHERE s.last_lsn >= (SELECT MAX(last_lsn)  FROM msdb.dbo.backupset WHERE [type] = 'D' and database_name = @dbName) " +
-                  "AND s.database_name = @dbName ORDER BY s.backup_start_date DESC, backup_finish_date";
+                  "WHERE s.last_lsn >= (" +
+                    "SELECT MAX(last_lsn) FROM msdb.dbo.backupset " +
+                    "WHERE [type] = 'D' " +
+                    "AND database_name = @dbName " +
+                    $"{(ignoreCopyOnlyBackups ? "AND is_copy_only = 0" : "")}" +
```
AdityaNPL, line 82: @alexirion10 from my understanding, I think there is an issue with that approach.
This part, I believe, determines how far in the past we are going to get the backups?
If so, we need to specify here if we are ignoring the `copy_only` backups or not.

Otherwise, we may have a situation like:
Let's say `A` is a "full" backup taken as part of the normal backup process and `B` is a "full" `copy_only` backup taken a bit later.
If we don't have the ignore logic here, the list will only contain backups till `B` but what we want is till `A`

alexirion10: ahh I see.  In that case I think this would be more simply always ignoring the copy only backups

AdityaNPL: don't we want the option for users to decide (and also not break the usage for users who already expect `copy_only`s)?

or is it the case that since our use-case for this module is very specific to getting the restore backup-chain that we should always ignore `copy_only` backups?

alexirion10: Do users already expect copy_only?

AdityaNPL: with the current implementation on `master`, their backup chains could have included `copy_only` backups 
(we didn't have a filter for it)

alexirion10: ya "could have" I get it. 
I'm asking: do we have any use for copy_only backups?  Do we ever need them in the backup chain?

AdityaNPL: Not for the prod-to-dev restore jobs, no
I cannot say for other use-cases where backup-chains are required too

AdityaNPL: sometimes you might want to create a `copy_only` backup chain for a specific purpose

alexirion10: I can't think of any purposes

AdityaNPL: It looks like the main purpose of a `copy_only` backup is to do a full restore from a specific moment in time 
(kind of like a pinned marker)
So a potential use case is when a `copy_only` "full" backup is taken at point `A`, and then after a few transactional backups, there was a full backup part of the normal backup process at point `B`.
The user might want to use the `copy_only` chain from `A` till just before `B` to restore a DB for a specific reason (like debugging an issue seen at that point)?
Inferred from:
https://dba.stackexchange.com/questions/45876/difference-between-full-backup-and-copy-only-full-backup
https://docs.microsoft.com/en-us/sql/relational-databases/backup-restore/copy-only-backups-sql-server

But for our module to support this, we will need a few more changes.
I agree, I can't see any purpose of getting a `copy_only` chain for backups taken after the most recent full backup
Will change it so that by design we ignore them 
(if we need to support `copy_only` DBs we can handle as another feature request)

alexirion10 (approved):

PR #62, opened 2021-03-01
fix(striped_backups): changed logic to get all striped backups

AdityaNPL:
Sometimes backups are split into multiple files (called `striped` backup), we need all of them for the restore process.
ref: https://www.sqlservercentral.com/articles/getting-a-list-of-the-striped-backup-files

### DIFF
--- a/src/BackupChain.cs
+++ b/src/BackupChain.cs
@@ -4,6 +4,7 @@ namespace AgDatabaseMove
   using System.Collections.Generic;
   using System.IO;
   using System.Linq;
+  using Exceptions;
   using SmoFacade;
 
 
@@ -19,26 +20,33 @@ namespace AgDatabaseMove
   {
     private readonly IList<BackupMetadata> _orderedBackups;
 
+    /// <summary>
+    /// Chain needs to be 1 full backup, 1 diff and then the logs in the order they were taken
+    /// Sometimes backups can be striped (i.e. split into multiple files) - so we need to handle these cases too
+    /// https://www.sqlservercentral.com/articles/getting-a-list-of-the-striped-backup-files
+    /// </summary>
+    /// <param name="recentBackups"></param>
     private BackupChain(IList<BackupMetadata> recentBackups)
     {
       var backups = recentBackups.Distinct(new BackupMetadataEqualityComparer())
-        .Where(b => IsValidFilePath(b)) // A third party application caused invalid path strings to be inserted into backupmediafamily
+        .Where(IsValidFilePath) // A third party application caused invalid path strings to be inserted into backupmediafamily
         .ToList();
 
-      var mostRecentFullBackup = MostRecentFullBackup(recentBackups);
-      _orderedBackups = new List<BackupMetadata> { mostRecentFullBackup };
+      var mostRecentFullBackups = MostRecentFullBackups(recentBackups).ToList();
+      
+      var differentialBackups = MostRecentDifferentialBackups(backups, mostRecentFullBackups.First());
+      // differentialBackups can be null
+      
+      var orderedBackups = mostRecentFullBackups.Concat(differentialBackups).ToList();
 
-      var differentialBackup = MostRecentDifferentialBackup(backups, mostRecentFullBackup);
-      if(differentialBackup != null)
-        _orderedBackups.Add(differentialBackup);
-
-      var mostRecentBackup = _orderedBackups.Last();
-      while(mostRecentBackup != null) {
-        mostRecentBackup = NextLogBackup(backups, mostRecentBackup);
-
-        if(mostRecentBackup != null)
-          _orderedBackups.Add(mostRecentBackup);
+      var prevBackup = orderedBackups.Last();
+      IEnumerable<BackupMetadata> nextLogBackups;
+      while((nextLogBackups = NextLogBackups(backups, prevBackup)).Any()) {
+        orderedBackups.AddRange(nextLogBackups);
+        prevBackup = orderedBackups.Last();
       }
+
+      _orderedBackups = orderedBackups;
     }
 
     /// <summary>
@@ -56,26 +64,40 @@ namespace AgDatabaseMove
     /// </summary>
     public IEnumerable<BackupMetadata> OrderedBackups => _orderedBackups;
 
-    private BackupMetadata MostRecentFullBackup(IList<BackupMetadata> backups)
+    private static IEnumerable<BackupMetadata> MostRecentFullBackups(IEnumerable<BackupMetadata> backups)
     {
-      return backups.Where(b => b.BackupType == BackupFileTools.BackupType.Full).OrderByDescending(d => d.CheckpointLsn)
-        .First();
+      var fullBackupsOrdered = backups
+        .Where(b => b.BackupType == BackupFileTools.BackupType.Full)
+        .OrderByDescending(d => d.CheckpointLsn).ToList();
+      if(!fullBackupsOrdered.Any()) {
+        throw new BackupChainException("Could not find any full backups");
+      }
+
+      var targetCheckpointLsn = fullBackupsOrdered.First().CheckpointLsn;
+      // get all the stripes of the most recent full backup (i.e. has the same CheckpointLsn)
+      return fullBackupsOrdered.Where(fullBackup => fullBackup.CheckpointLsn == targetCheckpointLsn); 
     }
 
-    private BackupMetadata MostRecentDifferentialBackup(IList<BackupMetadata> backups, BackupMetadata lastFullBackup)
+    private static IEnumerable<BackupMetadata> MostRecentDifferentialBackups(IEnumerable<BackupMetadata> backups, BackupMetadata lastFullBackup)
     {
-      return backups.Where(b => b.BackupType == BackupFileTools.BackupType.Diff &&
-                                b.DatabaseBackupLsn == lastFullBackup.CheckpointLsn)
-        .OrderByDescending(b => b.LastLsn).FirstOrDefault();
+      var diffBackupsOrdered = backups
+        .Where(b => b.BackupType == BackupFileTools.BackupType.Diff &&
+                    b.DatabaseBackupLsn == lastFullBackup.CheckpointLsn)
+        .OrderByDescending(b => b.LastLsn).ToList();
+
+      var targetLastLsn = diffBackupsOrdered.First().LastLsn;
+      // get all the stripes of the most recent diff backup (i.e. has the same LastLsn)
+      return diffBackupsOrdered.Where(diffBackup => diffBackup.LastLsn == targetLastLsn); 
     }
 
-    private BackupMetadata NextLogBackup(IList<BackupMetadata> backups, BackupMetadata prevBackup)
+    private static IEnumerable<BackupMetadata> NextLogBackups(IEnumerable<BackupMetadata> backups, BackupMetadata prevBackup)
     {
-      return backups.Where(b => b.BackupType == BackupFileTools.BackupType.Log)
-        .SingleOrDefault(d => prevBackup.LastLsn >= d.FirstLsn && prevBackup.LastLsn + 1 < d.LastLsn);
+      // get all the stripes of the next log backup
+      return backups.Where(b => b.BackupType == BackupFileTools.BackupType.Log && 
+                                prevBackup.LastLsn >= b.FirstLsn && prevBackup.LastLsn + 1 < b.LastLsn);
     }
 
-    private bool IsValidFilePath(BackupMetadata meta)
+    private static bool IsValidFilePath(BackupMetadata meta)
     {
       if(BackupFileTools.IsUrl(meta.PhysicalDeviceName))
         return true;
--- a/src/BackupChain.cs
+++ b/src/BackupChain.cs
@@ -76,7 +76,9 @@ namespace AgDatabaseMove
         .Where(b => b.BackupType == BackupFileTools.BackupType.Diff &&
                     b.DatabaseBackupLsn == lastFullBackup.CheckpointLsn)
         .OrderByDescending(b => b.LastLsn).ToList();
-
+      if (!diffBackupsOrdered.Any()) {
+        return new List<BackupMetadata>();
+      }
       var targetLastLsn = diffBackupsOrdered.First().LastLsn;
       // get all the stripes of this backup
       return diffBackupsOrdered.Where(diffBackup => diffBackup.LastLsn == targetLastLsn); 
--- a/src/BackupChain.cs
+++ b/src/BackupChain.cs
@@ -23,7 +23,7 @@ namespace AgDatabaseMove
     // This also handles any striped backups
     private BackupChain(IList<BackupMetadata> recentBackups)
     {
-      if (!(recentBackups?.Count > 0)) {
+      if (recentBackups == null || recentBackups.Count == 0) {
         throw new BackupChainException("There are no recent backups to form a chain");
       }
 
--- a/src/BackupChain.cs
+++ b/src/BackupChain.cs
@@ -32,7 +32,7 @@ namespace AgDatabaseMove
         .Where(IsValidFilePath) // A third party application caused invalid path strings to be inserted into backupmediafamily
         .ToList();
 
-      var mostRecentFullBackups = MostRecentFullBackups(recentBackups).ToList();
+      var mostRecentFullBackups = MostRecentFullBackups(backups).ToList();
       
       var differentialBackups = MostRecentDifferentialBackups(backups, mostRecentFullBackups.First());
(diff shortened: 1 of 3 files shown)


src/BackupChain.cs
```diff
@@ -19,26 +20,33 @@ public class BackupChain : IBackupChain
...
+      var orderedBackups = mostRecentFullBackups.Concat(differentialBackups).ToList();
+
+      var prevBackup = orderedBackups.Last();
+      IEnumerable<BackupMetadata> nextLogBackups;
+      while((nextLogBackups = NextLogBackups(backups, prevBackup)).Any()) {
+        orderedBackups.AddRange(nextLogBackups);
+        prevBackup = orderedBackups.Last();
       }
+
+      _orderedBackups = orderedBackups;
```
alexirion10, line 49: I wonder if it's not easier to just get all of the backup files matching the `LastFullBackup.DatabaseBackupLsn`, then ordering all of them on LastLsn/name at the end

AdityaNPL: Just to clarify, you mean get the **`LastFullBackup.CheckpointLsn`** and then match all backups that have that same value in `backup.DatabaseBackupLsn`?

hmm that could work, but we would need to also filter out any transaction/diff backups that happened before the most recent diff backup - but that should be a simple LINQ query or List remove

I'll have a look at doing that

alexirion10: ahhh yeah that's why we chain.... nevermind don't do that

src/BackupChain.cs
```diff
@@ -19,26 +20,33 @@ public class BackupChain : IBackupChain
   {
     private readonly IList<BackupMetadata> _orderedBackups;
 
+    /// <summary>
+    /// Chain needs to be 1 full backup, 1 diff and then the logs in the order they were taken
+    /// Sometimes backups can be striped (i.e. split into multiple files) - so we need to handle these cases too
+    /// https://www.sqlservercentral.com/articles/getting-a-list-of-the-striped-backup-files
+    /// </summary>
```
alexirion10, line 27: You don't need to doc comment on every method or comment each line of code. 
My viewpoint on this is that if a function or code requires a doc comment to explain what it's doing, then the function is too complicated and needs to be broken up.

src/BackupChain.cs
```diff
@@ -56,26 +64,40 @@ private BackupChain(IList<BackupMetadata> recentBackups)
...
+                    b.DatabaseBackupLsn == lastFullBackup.CheckpointLsn)
+        .OrderByDescending(b => b.LastLsn).ToList();
+
+      var targetLastLsn = diffBackupsOrdered.First().LastLsn;
+      // get all the stripes of the most recent diff backup (i.e. has the same LastLsn)
+      return diffBackupsOrdered.Where(diffBackup => diffBackup.LastLsn == targetLastLsn); 
     }
 
-    private BackupMetadata NextLogBackup(IList<BackupMetadata> backups, BackupMetadata prevBackup)
+    private static IEnumerable<BackupMetadata> NextLogBackups(IEnumerable<BackupMetadata> backups, BackupMetadata prevBackup)
```
alexirion10, line 93: Can you keep the method names singular?  NextLogBackups sounds like you are returning multiple backups when it's 1 backup, striped into several files

src/BackupChain.cs
```diff
@@ -19,26 +20,33 @@ public class BackupChain : IBackupChain
...
-        mostRecentBackup = NextLogBackup(backups, mostRecentBackup);
-
-        if(mostRecentBackup != null)
-          _orderedBackups.Add(mostRecentBackup);
+      var mostRecentFullBackups = MostRecentFullBackups(recentBackups).ToList();
+      
+      var differentialBackups = MostRecentDifferentialBackups(backups, mostRecentFullBackups.First());
+      // differentialBackups can be null
+      
+      var orderedBackups = mostRecentFullBackups.Concat(differentialBackups).ToList();
```
alexirion10, line 40: Why'd you change this logic?
You can use `AddRange` instead of `Add`

alexirion10 (commented):
I think the logic in here is right... but this needs to be tested more thouroughly

src/BackupChain.cs
AdityaNPL: ahh sorry, residual from when I was using `IEnumerable` everywhere
Will change this

src/BackupChain.cs
AdityaNPL: This was more for a notice that this function can also handle striped backups in the chain
Will make it a single line comment and shorter just to state that point

AdityaNPL:
Sure, I have tested locally with a MSSQL instance that has striped backups - worked correctly

Will also try to add some negative tests by adding duplicates and diff/full backups with diff LSNs

alexirion10 (approved):

src/BackupChain.cs
```diff
@@ -19,26 +20,28 @@ public class BackupChain : IBackupChain
   {
     private readonly IList<BackupMetadata> _orderedBackups;
 
+    // This also handles any striped backups
     private BackupChain(IList<BackupMetadata> recentBackups)
     {
+      if (recentBackups == null || !recentBackups.Any()) {
```
alexirion10, line 26: `Empty` would be the appropriate check here instead of `any`

alexirion10: What if a full backup doesn't exist?  Should you throw there too? What if a log backup doesn't exist?

AdityaNPL: Don't think there is `.Empty()` for `List` - preferred strategy by others seem to be this too:
https://stackoverflow.com/questions/24390005/checking-for-empty-or-null-liststring/43142462

I throw an exception in the function that searches for a full backup: 
https://github.com/factset/AgDatabaseMove/pull/62/files#diff-e1d6a44c77500bb8e79c7f8f9f446183e74633feb51f39573323c6345b8f6bf1R69

don't need one for `diff` or `logs` (they don't necessarily have to be in the chain for the restore to work)
can be `full -> diff -> log`, `full -> diff`, `full -> Log` or just `full`
(the order matters tho, `full` before `diff` and `diff` before any `logs` - creating some new tests that will ensure this)

alexirion10: Gotcha. I think `Count >0` is slightly more readable.
Ah I think in CRDS jobs, we force a new log backup so there is always at least 1 log, but I suppose that isn't always required

tests/AgDatabaseMove.Unit/BackupOrder.cs
```diff
@@ -44,51 +60,158 @@ private static List<BackupMetadata> ListBackups()
...
+                      prevBackup.LastLsn == b.FirstLsn)
+                    );
+        prevBackup = b;
+      });
 
-      Assert.Equal<IEnumerable>(backupChain.OrderedBackups, expected);
     }
 
-    [Fact]
-    public void MissingLink()
```
AdityaNPL, line 84: @alexirion10 not sure how to test this - any ideas?
I don't think the current testing strategy for this is a complete one - i.e just asserting that the length of the output is not the length of the expected output 

I think we should ideally have a way to detect this in our logic and throw an exception if it occurs - then test for the exception here.

One potential idea I have is to see if there were other log backups still left in the list we didn't include in the chain - and go through their LSNs to see if they should have been in the chain - hence identifying the missing link

But is a bit complicated to do - so I'm wondering if it's best for another PR?
What are your thoughts?

alexirion10: no no you don't need to do all of that.  The existing test is a good one

1. Get a full backup chain.
2. Remove 1 of the items from the middle of the chain
Assert 1 has all backups in the chain up to the one removed.  

We rely on the SQL tech for chaining LSN's. Once the chain is broken, I'm not aware of how to find a missing link.

tests/AgDatabaseMove.Unit/BackupOrder.cs
```diff
@@ -44,51 +60,158 @@ private static List<BackupMetadata> ListBackups()
...
+                                     b.CheckpointLsn == full.CheckpointLsn);
+
+      var otherFullBackups = backupChainList.RemoveAll(b => b.BackupType == BackupFileTools.BackupType.Full);
+      Assert.Equal(0, otherFullBackups);
+
+      // next one can be diff or log
+      var currentBackup = backupChainList.FirstOrDefault();
+      if(currentBackup == null) {
+        return;
+      }
```
alexirion10, line 157: I get that this works and all.... but it seems like you could reduce this 50 line function into a 5 line do()while() loop checking that lastLsn == next.FirstLsn for the entire chain

tests/AgDatabaseMove.Unit/BackupOrder.cs
AdityaNPL: ahh I see, sure will place that test back in

>We rely on the SQL tech for chaining LSN's

I guess this is also to do with scope right - this module assumes LSNs are done correctly?

tests/AgDatabaseMove.Unit/BackupOrder.cs
AdityaNPL: yes, so that is pretty much what I have done to check the log backups:
https://github.com/factset/AgDatabaseMove/blob/8abb6abb67b5a0ce5a3c69239d1dcb86df910c5d/tests/AgDatabaseMove.Unit/BackupOrder.cs#L171-L179

But all of this before is to make sure:
- The first backup in the chain is a full backup
- There are no other full backups (excluding the stripes of the first one)
- If there is a diff backup, then it is the only one and that its backupLsn == full.checkpointLsn
- And this will also make sure that the order is maintained full then diff (if any) then logs (if any) 

But I do see your point, looking at this now maybe there is a way to refactor this to be more simple
Let me give it a try

AdityaNPL: Refactored it to be a bit shorter - let me know what you think

tests/AgDatabaseMove.Unit/BackupOrder.cs
```diff
@@ -44,51 +60,161 @@ private static List<BackupMetadata> ListBackups()
...
+    {
+      var list = GetBackupList();
+      list.RemoveAll(b => b.BackupType == BackupFileTools.BackupType.Full);
+      return list;
+    }
+
+    private static void VerifyListIsAValidBackupChain(List<BackupMetadata> backupChain)
+    {
+      var listOfFileNames = backupChain.Select(b => b.PhysicalDeviceName).ToList();
+      Assert.True(listOfFileNames.Count == listOfFileNames.Distinct().Count(), "There are no duplicates?");
```
alexirion10, line 131: `Assert.Equal` instead of `Assert.True(x == y)`
Is this an assertion or a question?  Why are you testing for distinct file names?  I don't think that really provides much value here

tests/AgDatabaseMove.Unit/BackupOrder.cs
```diff
@@ -44,51 +60,161 @@ private static List<BackupMetadata> ListBackups()
...
+
+      var fullBackup = backupChain.First();
+      Assert.NotNull(fullBackup);
+      Assert.True(fullBackup.BackupType == BackupFileTools.BackupType.Full, "Is the first backup a full backup?");
```
alexirion10, line 135: `Equals` instead of `True(x==y)`. No need for the comment string, this should be obvious.
I see the same thing all over this file, so I'm not going to comment on each one

tests/AgDatabaseMove.Unit/BackupOrder.cs
```diff
@@ -44,51 +60,161 @@ private static List<BackupMetadata> ListBackups()
...
+      while((currentBackup = backupChain.FirstOrDefault())?.BackupType == BackupFileTools.BackupType.Log) {
+        Assert.True(currentBackup.DatabaseBackupLsn == fullBackup.CheckpointLsn,
+                    "Is the log linked to the full backup?");
+        Assert.True(
+                    (currentBackup.FirstLsn == prevBackup.FirstLsn && currentBackup.LastLsn == prevBackup.LastLsn) ||
+                    (prevBackup.LastLsn == currentBackup.FirstLsn),
+                    "Is it either the striped file of the prev log backup or the next one in the chain?");
+
+        prevBackup = currentBackup;
+        backupChain.RemoveAt(0);
```
alexirion10, line 168: This test has 3 while loops... I think you can simplify it down into 1 with some booleans to help:
```C#
bool foundFull, foundDiff, foundLog = false;
while(curr = backupChain.FirstOrDefault())
{
  if(curr.BackupType == Full)
  {
    if (foundDiff || foundLog)
       Assert.False():
    foundFull = true;
    backupChain.Remove(b => b.CheckPointLSN == curr.CheckpointLSN)
  }
  if(curr.BackupType == Diff)
   {
       if (foundDiff || foundLog || !foundFull)
         Assert.False():
      backupChain.Remove(b => b.LastLSN== curr.LastLSN)
   }
  if(curr.BackupType == Log)
  {
    if (!foundFull)
         Assert.False():
    backupChain.Remove(b => b.LastLSN== curr.LastLSN)
  }
}


```

AdityaNPL: Ah yes, thank you - I think I was stuck in a pattern of thinking we need to handle each in sequence
Rather we can handle it more like "when we get this type -> Assert this, this and that" 

This almost works - just need to add in the ordering/link checks for the logs and diff (i.e. they all point to the `full.checkpointLsn` and that for each `prev.lastLsn==curr.firstLns`)
Will implement it!

tests/AgDatabaseMove.Unit/BackupOrder.cs
AdityaNPL: It is an Assertion - will change all those types in the file to `.Equal()`

I put this in cause of the comment that was here:
https://github.com/factset/AgDatabaseMove/blob/683f409bf8d9565af2ec824eb84ce923225b0a4b/src/BackupMetadata.cs#L8-L13

alexirion10: gotcha.  
That should be an entirely separate and independent test then => a new test method that creates a list of backups (two with the same filename) and tests that the duplicate names get removed on select distinct.
No need for this assertion to be in the logic checking for a valid LSN chain

AdityaNPL: ah I see - agreed!
Will change

tests/AgDatabaseMove.Unit/BackupOrder.cs
```diff
@@ -44,51 +60,155 @@ private static List<BackupMetadata> ListBackups()
...
+    }
+
+    [Fact]
+    public void DuplicateFiles()
+    {
+      var backups = GetBackupListWithStripesAndDuplicates();
+      var agDatabase = new Mock<IAgDatabase>();
+      agDatabase.Setup(agd => agd.RecentBackups()).Returns(backups);
+      var chain = new BackupChain(agDatabase.Object).OrderedBackups.ToList();
+      Assert.Equal(chain.Count, backups.Count/2);
```
alexirion10, line 211: `Assert.Equal(chain.Count, backups..GroupBy(b =>b.FirstLsn).Count);`

src/BackupChain.cs
```diff
@@ -19,26 +20,28 @@ public class BackupChain : IBackupChain
   {
     private readonly IList<BackupMetadata> _orderedBackups;
 
+    // This also handles any striped backups
     private BackupChain(IList<BackupMetadata> recentBackups)
     {
+      if (!(recentBackups?.Count > 0)) {
+        throw new BackupChainException("There are no recent backups to form a chain");
+      }
```
alexirion10, line 28: First off, if inputs are possibly null, then you should check that before all else.
Second - there is too many operators here: `if(recentBackups.empty())`

alexirion10 (approved):
two small fixes, then good to go

tests/AgDatabaseMove.Unit/BackupOrder.cs
AdityaNPL: ah I see - I did the direct compare cause I know what the backup list generator for this test spits back out (duplicates each file)
but you are right here, should be using a more general logic for the check in case we change the generator

src/BackupChain.cs
AdityaNPL: `IList` doesn't have `Empty()`
https://github.com/factset/AgDatabaseMove/pull/62#discussion_r585817633

sure, will change to:
`if (recentBackups == null || recentBackups.Count == 0) { throw ... }`

tests/AgDatabaseMove.Unit/BackupOrder.cs
AdityaNPL: (just to note, striped backups have same LSNs so will use file name in group by clause)

PR #64, opened 2021-03-29
fix(IsValidFilePath): Can now validate Windows UNC paths on Unix runtime

AdityaNPL:
Before, `Path.IsPathRooted(meta.PhysicalDeviceName);` would return false for Windows UNC paths on Unix runtime due to a difference in dotnet's implementation of the function for the diff runtimes:

Windows Runtime: https://github.com/dotnet/runtime/blob/44f8f0faee42367bdd39277d8295f6e914ad7f4a/src/libraries/System.Private.CoreLib/src/System/IO/Path.Windows.cs#L194 
vs.
Unix Runtime: https://github.com/dotnet/runtime/blob/44f8f0faee42367bdd39277d8295f6e914ad7f4a/src/libraries/System.Private.CoreLib/src/System/IO/Path.Unix.cs#L113

This would lead to backup files being filtered out and an exception being thrown during the creation of the backup chain.

Now we use the `IsUnc` property of the `Uri` class to detect UNC paths and since UNCs have to be fully-qualified by design, we can validate them straight away (just like how we do for URLs).

Have also made a few refactors to ease testing of this new approach.

### DIFF
--- a/src/BackupChain.cs
+++ b/src/BackupChain.cs
@@ -99,7 +99,7 @@ namespace AgDatabaseMove
     private static bool IsValidFilePath(BackupMetadata meta)
     {
       var path = meta.PhysicalDeviceName;
-      return BackupFileTools.IsUrl(path) || BackupFileTools.IsValidPath(path);
+      return BackupFileTools.IsValidFileUrl(path) || BackupFileTools.IsValidFilePath(path);
     }
   }
 }
--- a/src/BackupChain.cs
+++ b/src/BackupChain.cs
@@ -98,7 +98,8 @@ namespace AgDatabaseMove
 
     private static bool IsValidFilePath(BackupMetadata meta)
     {
-      return BackupFileTools.IsUrl(meta.PhysicalDeviceName) || BackupFileTools.IsValidPath(meta.PhysicalDeviceName);
+      var path = meta.PhysicalDeviceName;
+      return BackupFileTools.IsUrl(path) || BackupFileTools.IsUnc(path) || BackupFileTools.IsValidPath(path);
     }
   }
 }
--- a/src/BackupChain.cs
+++ b/src/BackupChain.cs
@@ -99,7 +99,7 @@ namespace AgDatabaseMove
     private static bool IsValidFilePath(BackupMetadata meta)
     {
       var path = meta.PhysicalDeviceName;
-      return BackupFileTools.IsUrl(path) || BackupFileTools.IsUnc(path) || BackupFileTools.IsValidPath(path);
+      return BackupFileTools.IsUrl(path) || BackupFileTools.IsValidPath(path);
     }
   }
 }
--- a/src/BackupChain.cs
+++ b/src/BackupChain.cs
@@ -98,22 +98,7 @@ namespace AgDatabaseMove
 
     private static bool IsValidFilePath(BackupMetadata meta)
     {
-      if(BackupFileTools.IsUrl(meta.PhysicalDeviceName))
-        return true;
-
-      // A quick check before leaning on exceptions
-      if(Path.GetInvalidPathChars().Any(meta.PhysicalDeviceName.Contains))
-        return false;
-
-      try {
-        // This will throw an argument exception if the path is invalid
-        Path.GetFullPath(meta.PhysicalDeviceName);
-        // A relative path won't help us much if the destination is another server. It needs to be rooted.
-        return Path.IsPathRooted(meta.PhysicalDeviceName);
-      }
-      catch(Exception) {
-        return false;
-      }
+      return BackupFileTools.IsUrl(meta.PhysicalDeviceName) || BackupFileTools.IsValidPath(meta.PhysicalDeviceName);
     }
   }
 }
--- a/src/SmoFacade/BackupFileTools.cs
+++ b/src/SmoFacade/BackupFileTools.cs
@@ -15,11 +15,11 @@
       Log // trn  / L
     }
 
-    public static bool IsUrl(string path)
+    public static bool IsValidFileUrl(string path)
     {
       return Uri.TryCreate(path, UriKind.Absolute, out var uriResult)
              && (uriResult.Scheme == Uri.UriSchemeHttp || uriResult.Scheme == Uri.UriSchemeHttps || uriResult.IsUnc)
-             && !(path.EndsWith(@"/") || path.EndsWith(@"\"));
+             && Path.HasExtension(path);
     }
 
     public static string BackupTypeToExtension(BackupType type)
@@ -50,7 +50,7 @@
       }
     }
 
-    public static bool IsValidPath(string path)
+    public static bool IsValidFilePath(string path)
     {
       // A quick check before leaning on exceptions
       if(Path.GetInvalidPathChars().Any(path.Contains)) {
@@ -61,7 +61,7 @@
         // This will throw an argument exception if the path is invalid
         Path.GetFullPath(path);
         // A relative path won't help us much if the destination is another server. It needs to be rooted.
-        return Path.IsPathRooted(path);
+        return Path.IsPathRooted(path) && Path.HasExtension(path);
       }
       catch(Exception) {
         return false;
--- a/src/SmoFacade/BackupFileTools.cs
+++ b/src/SmoFacade/BackupFileTools.cs
@@ -1,6 +1,8 @@
 ﻿namespace AgDatabaseMove.SmoFacade
 {
   using System;
+  using System.IO;
+  using System.Linq;
   using System.Text.RegularExpressions;
 
 
@@ -45,5 +47,23 @@
           throw new ArgumentException("Invalid backup type");
       }
     }
+
+    public static bool IsValidPath(string path)
+    {
+      // A quick check before leaning on exceptions
+      if(Path.GetInvalidPathChars().Any(path.Contains))
+        return false;
+
+      try {
+        // This will throw an argument exception if the path is invalid
+        Path.GetFullPath(path);
+        // A relative path won't help us much if the destination is another server. It needs to be rooted.
+        return Path.IsPathRooted(path) || path.StartsWith(@"\\");
+      }
+      catch(Exception) {
+        return false;
+      }
+    }
+
   }
 }
--- a/src/SmoFacade/BackupFileTools.cs
+++ b/src/SmoFacade/BackupFileTools.cs
@@ -51,8 +51,7 @@
     public static bool IsValidPath(string path)
     {
       // A quick check before leaning on exceptions
-      if (Path.GetInvalidPathChars().Any(path.Contains))
-      {
+      if(Path.GetInvalidPathChars().Any(path.Contains)) {
         return false;
       }
 
--- a/src/SmoFacade/BackupFileTools.cs
+++ b/src/SmoFacade/BackupFileTools.cs
@@ -17,7 +17,9 @@
 
     public static bool IsUrl(string path)
     {
-      return Regex.IsMatch(path, @"(http(|s):\/)(\/[^\s]+)+\.([a-zA-Z]+)$");
+      return Uri.TryCreate(path, UriKind.Absolute, out var uriResult)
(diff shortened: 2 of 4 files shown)


AdityaNPL:
@alexirion10 @Kaneraz 
Wondering if we can use `Uri` for this function too and simplify it so that the UNC check goes there instead?

https://github.com/factset/AgDatabaseMove/blob/babf28a49efb00e4bf4385baf8c1fe7b2184db21/src/SmoFacade/BackupFileTools.cs#L16-L19

alexirion10 (commented):
Looks good.  Ya you should be able to use `new Uri` constructor on URLs instead of the regex.
I thought we were done with DFS backup paths by now...

AdityaNPL:
Cool, will change that function too.
We are migrating away from it, but looks like we do need to support some DBs with it for the time being.

src/SmoFacade/BackupFileTools.cs
```diff
@@ -15,7 +17,9 @@ public enum BackupType
 
     public static bool IsUrl(string path)
     {
-      return Regex.IsMatch(path, @"(http(|s):\/)(\/[^\s]+)+\.([a-zA-Z]+)$");
+      return Uri.TryCreate(path, UriKind.Absolute, out var uriResult)
+             && (uriResult.Scheme == Uri.UriSchemeHttp || uriResult.Scheme == Uri.UriSchemeHttps || uriResult.IsUnc)
+             && !(path.EndsWith(@"/") || path.EndsWith(@"\"));
```
AdityaNPL, line 22: had to make a few tweaks to get the same functionality
tests confirm that it is working

src/SmoFacade/BackupFileTools.cs
```diff
@@ -17,7 +17,9 @@ public enum BackupType
...
```
alexirion10, line 22: This function does much more than the signature `IsUrl` would indicate.
I'd rename to `IsValidUriFile`

Also, the logic `&& !(path.EndsWith(@"/") || path.EndsWith(@"\"))` can be improved upon.  
You omitted the test below for the case `http://storage-account.blob.core.windows.net/container` which doesn't have a trailing slash.
That case would fail this test here.
I think using a regex or [Ssytem.IO.Path library](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.hasextension?view=net-5.0) to validate the path has an extension here would be more ideal

tests/AgDatabaseMove.Unit/FileToolsTest.cs
```diff
@@ -7,38 +7,35 @@ namespace AgDatabaseMove.Unit
...
-      new object[]
-        { "https://storage-account.blob.core.windows.net/container/sql/db_name/backup_2020_09_02_170003_697.trn" },
-      new object[] { "http://hello/a.bak" }
-    };
-
-    public static IEnumerable<object[]> NonUrlFileExamples => new List<object[]> {
-      new object[] { @"c:\hello\a.bak" },
-      new object[] { @"\\abc\hello/a.bak" },
-      new object[] { "https://storage-account.blob.core.windows.net/container" },
-      new object[] { "http://storage-account.blob.core.windows.net/container" }
```
alexirion10, line 27: you aren't using these cases anymore. I'm pretty sure the last two here would break your tests :p
Add them back please

alexirion10 (changes requested):

src/SmoFacade/BackupFileTools.cs
AdityaNPL: Sure will change the name to something more reflective of the implementation.

The issue I found here is that not all files have an extension. That is why I also changed the test cases to reflect this.
Although I think I just had one positive test case there that showed this and that too only for UNC (`[InlineData(@"\\server\file")]`) - so can add some more there?

If we do want to restrict to just files with extensions I think we can do something like:
`new Uri(url).AbsolutePath.Split('/').Last().Contains('.');` or maybe just check for `Path.GetExtension(url)`
https://stackoverflow.com/a/53262979/15521092

A more concrete check would be to see if the file actually exists - `File.Exists(url)`
But in some cases the consumer of this module might not have access to the files themselves (I am facing this now) - so I think we should stick to just checking the syntax of the url.

Please let me know your thoughts.

tests/AgDatabaseMove.Unit/FileToolsTest.cs
AdityaNPL: Will add back the test case for `c:\hello\a.bak` test case - that still works

For `\\abc\hello/a.bak` - unc urls are a valid URL here
_Quick note:_ URI class is a bit more intelligent and flips the slashes
This [article](https://docs.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/5.0/unc-path-recognition-unix) explains it on .Net 5, but looks like they have something similar for .Net core 2.0 too - [docs](https://docs.microsoft.com/en-us/dotnet/api/system.uri.localpath?view=netcore-2.0#remarks)

For `"https://storage-account.blob.core.windows.net/container"` and `"http://storage-account.blob.core.windows.net/container"` - pls see comment above - if needed, I'll add them back

AdityaNPL:
@alexirion10 I'm just wondering now, do we really need all these checks for the url/path.
Can't we just use it as it is and let the consumer deal with any issues that arise from it (at the end of the day the issue was because of the metadata we got from their MSSQL instance)? 
I don't see why this module should be in charge of validating them?

I believe invalid, relative and incorrect paths/urls would error out naturally when you try to use it?

Moreover, I believe having these silent filters here could lead to a head scratching problem for users as their files might mysteriously disappear in the backup chain without any warnings in some cases.

What's your view on this?

src/SmoFacade/BackupFileTools.cs
alexirion10: What files don't have extensions?  Surely all of our SQL backup files do.  Consult with Angelo and let me know if that assumption is wrong for some reason.

`System.IO.Path.HasExtension` is all that should be needed.  https://docs.microsoft.com/en-us/dotnet/api/system.io.path.hasextension?view=net-5.0

`File.Exists()` isn't going to work for URL file paths

tests/AgDatabaseMove.Unit/FileToolsTest.cs
alexirion10: This comment only meant to discuss line 26 & 27, I think the first two still existed.
Correct, the last two are not valid files

AdityaNPL:
My opinion here is that we can treat our functions in this module (like  `_options.Destination.Restore`) as consumers of the backup chain too. So we let it handle any issues with incorrect paths however it wants to. 
(basically the `BackupChain` class takes no responsibility for checking if meta-data is correct, just returns it as is but in the right order)

I believe, this will keep things loosely-coupled and we won't have to handle edge-cases like these in the future.
Let me know what you think.

---

To address this particular example:
https://github.com/factset/AgDatabaseMove/blob/babf28a49efb00e4bf4385baf8c1fe7b2184db21/src/AgDatabaseMove.cs#L76

looks like under the hood it calls this SMO function: `Microsoft.SqlServer.Management.Smo.Restore.SqlRestore(Server)`:
https://github.com/factset/AgDatabaseMove/blob/babf28a49efb00e4bf4385baf8c1fe7b2184db21/src/SmoFacade/Server.cs#L164

So I was thinking a better design would be to let the SMO function raise an exception during the restore process if it can't get the file for any reason. We can then propagate the error back to the user.  
(same for any other function, we let the underlying tool throw the error when it can't get the file and we propagate it back)

src/SmoFacade/BackupFileTools.cs
AdityaNPL: Ah you are right. I was thinking of files in general (filed don't need extensions to work - [ref](https://superuser.com/questions/731323/does-one-really-need-to-use-file-extensions-for-files-stored-on-their-hard-drive#:~:text=Short%20answer%3A%20No%2C%20we%20don,then%20double%20click%20on%20it.))

But maybe for SQL backup files we will always have an extension like `.trn`/`.full`/`.diff` etc..
Will double check and get back

am99: It is safe to assume that all backup files will have extensions.  This will definitely be true for any backups created by the standard SQL backup jobs, and that's all we have to care about.  Even manual backups typically are created with an extension, though not necessarily the same as those created by our jobs (.full, .diff, .trn).  Yes it is technically possible for someone to come in and take a manual backup that does not have an extension, but we can say those backups won't be supported.

alexirion10:
You're then going to need to check file validity in every consuming location, hence duplicating that code and logic everywhere.  I don't want to go down that path.
SMO.restore is already going to throw if the file doesn't exist or it can't restore for whatever reason.
This "tightly coupled" file validity check is just to assure us that the _path_ is somewhat reasonable, not that it'll reachable or that it's actually a SQL backup file

AdityaNPL:
>So we let it handle any issues with incorrect paths however it wants to.

Just re-read my comment above and looks like it was a bit confusing here.
What I mean to say is that consumers (i.e. end-users of this tool) can then handle errors with wrong paths in any way they want. 

For the functions in this tool that consume the backup chain, it can just return the same error it gets from the underlying SMO function - so no need to duplicate any logic anywhere.

src/SmoFacade/BackupFileTools.cs
alexirion10: Noting I haven't re-reviewed because this hasn't been changed yet

AdityaNPL:
Had a conversation with @alexirion10 offline about this.

We came to the conclusion that there may be some aspects of `AgDatabaseMove` module that may restrict us in getting rid of the path/url checks all together right now. Mainly the error recovery design if the restore process fails because of incorrect paths (it might not be re-runnable).

For this PR we decided to keep it simple and just change the url/path checks and the respective tests. 

However, I want to see if we can remove these checks easily. 
Will need some investigating before we decide, so filed an issue to be placed in the backlog to be looked at a later time - #65

src/SmoFacade/BackupFileTools.cs
AdityaNPL: Done!

tests/AgDatabaseMove.Unit/FileToolsTest.cs
AdityaNPL: moved them back as invalid test data - tests pass

src/BackupChain.cs
```diff
@@ -98,22 +98,8 @@ private static IEnumerable<BackupMetadata> NextLogBackup(IEnumerable<BackupMetad
...
-        // This will throw an argument exception if the path is invalid
-        Path.GetFullPath(meta.PhysicalDeviceName);
-        // A relative path won't help us much if the destination is another server. It needs to be rooted.
-        return Path.IsPathRooted(meta.PhysicalDeviceName);
-      }
-      catch(Exception) {
-        return false;
-      }
+      var path = meta.PhysicalDeviceName;
+      return BackupFileTools.IsValidFileUrl(path) || BackupFileTools.IsValidFilePath(path);
```
alexirion10, line 102: Does `IsValidFilePath` ever return true here when `IsValidFileUrl` returns false?
I thought the `IsUnc` check took care of the filepath check

AdityaNPL: yes, `IsValidFilePath()` now only validates syntax for local files like these:
- `/some/file.ext` 
- `C:\dir\file.ext`

^^ but `IsValidFileUrl()` (which now contains the UNC check) would return false for these 

I'm assuming it's cause this tool could be running on the same box where backups are stored?
Also another usecase for `IsValidFileUrl()` is here to distinguish the type of path:
- https://github.com/factset/AgDatabaseMove/pull/64/files#diff-b26b234eb55bcea6a1c5cd567db2b6f7e7db565e30a58c70ac62cb22ddf722b1R135
- https://github.com/factset/AgDatabaseMove/pull/64/files#diff-b26b234eb55bcea6a1c5cd567db2b6f7e7db565e30a58c70ac62cb22ddf722b1R210

AdityaNPL: I've left the logic the same as what it was before for `IsValidFilePath()` (except for adding a check for file extensions)

alexirion10 (approved):

PR #66, opened 2021-05-07
fix(restore): polly retry on Azure blob storage single file handle issue

alexirion10:
Azure blob storage single filehandle only, concurrent restores will succeed with retry logic

### DIFF
--- a/src/AgDatabase.cs
+++ b/src/AgDatabase.cs
@@ -26,7 +26,8 @@ namespace AgDatabaseMove
     List<BackupMetadata> RecentBackups();
     void JoinAg();
 
-    void Restore(IEnumerable<BackupMetadata> backupOrder, Func<string, string> fileRelocation = null);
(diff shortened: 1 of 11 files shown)


Kaneraz (commented):
I like the addition of the policy, but we should get the retry duration a configurable value. Is my math right saying it's 13 min atm? We may want to stretch that out based on the database size.

Also, lots of formatting changes in here. Make sure they're in line with the jetbrains rules. Looks like it had issues running code cleanup in the CI build
```Failed to run cleanup. Unable to find the code cleanup profile with 'AgDatabaseMove' name```

alexirion10:
@Kaneraz Correct, I have 10+20+30+40+50+60+70+80+90+100+110+120 = 13 mins total.
Configurable as in pass a Polly policy into the AgDatabaseMove class?

Ya the Azure pipeline wasn't reporting status back to github PR at first... I ran the code cleanup locally.  Not sure what got out of sync in this pipeline

alexirion10:
@Kaneraz - can you re-review please?

Kaneraz (approved):

Issue #46, opened 2020-10-22
Check disk space before move


alexirion10:
check destination cluster disk space prior to the move.
Error out of the program before attempting to move if it'll cause the available disk to drop below 10% available

alexirion10:
https://github.com/factset/AgDatabaseMove/issues/46

PR #68, opened 2021-06-03
feat(database): size

alexirion10:
return database size
return server remaining disk size

tested on a few of our inhouse / QA boxes

### DIFF
--- a/src/SmoFacade/Server.cs
+++ b/src/SmoFacade/Server.cs
@@ -61,10 +61,11 @@ namespace AgDatabaseMove.SmoFacade
     {
       var query = "SELECT total_size_mb = (SUM(size) * 8. / 1024) " +
                   "FROM sys.master_files WITH(NOWAIT) " +
-                  $"WHERE database_id = DB_ID('{dbName}') " +
+                  "WHERE database_id = DB_ID(@db_name) " +
                   "GROUP BY database_id";
 
       using var cmd = SqlConnection.CreateCommand();
+      cmd.Parameters.Add(new SqlParameter("db_name", SqlDbType.VarChar) { Value = dbName });
       cmd.CommandText = query;
       using var reader = cmd.ExecuteReader();
       reader.Read();
--- a/src/SmoFacade/Server.cs
+++ b/src/SmoFacade/Server.cs
@@ -59,10 +59,10 @@ namespace AgDatabaseMove.SmoFacade
 
     public decimal DatabaseSizeMb(string dbName)
     {
-      string query = "SELECT total_size_mb = (SUM(size) * 8. / 1024) " + 
-                     "FROM sys.master_files WITH(NOWAIT) " +
-                     $"WHERE database_id = DB_ID('{dbName}') " + 
-                     "GROUP BY database_id";
+      var query = "SELECT total_size_mb = (SUM(size) * 8. / 1024) " +
+                  "FROM sys.master_files WITH(NOWAIT) " +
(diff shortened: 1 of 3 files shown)


src/SmoFacade/Server.cs
```diff
@@ -57,6 +57,34 @@ public void DropLogin(LoginProperties login)
       _server.Logins[login.Name].DropIfExists();
     }
 
+    public decimal DatabaseSizeMb(string dbName)
+    {
+      string query = "SELECT total_size_mb = (SUM(size) * 8. / 1024) " + 
+                     "FROM sys.master_files WITH(NOWAIT) " +
```
Kaneraz, line 63: Nice, this includes log size? That's a good metric to use.

src/SmoFacade/Server.cs
```diff
@@ -57,6 +57,34 @@ public void DropLogin(LoginProperties login)
...
+      reader.Read();
+      return (decimal)reader["total_size_mb"];
+    }
+
+    public int RemainingDiskMb()
+    {
+      string query = "SELECT DISTINCT CONVERT(INT, dovs.available_bytes/1048576.0) AS free_size_mb " +
+                     "FROM sys.master_files mf " +
+                     "CROSS APPLY sys.dm_os_volume_stats(mf.database_id, mf.FILE_ID) dovs " +
+                     "WHERE dovs.logical_volume_name = 'Data'";
```
Kaneraz, line 79: Is logical_volume_name always data? or is that just convention?

src/SmoFacade/Server.cs
alexirion10: correct, the `SUM()` aggregate on the `group by` combines log and data size

src/SmoFacade/Server.cs
alexirion10: Not sure what you're asking here.  Other values for `logical_volume_name` include TempDB, which is mounted on the T:\ volume in our case.
The query above aggregates everything on the D:\ drive which includes log and data files

alexirion10: discussed with Kaneraz offline, updated query to match on SERVERPROPERTY('InstanceDefaultDataPath')

src/SmoFacade/Server.cs
```diff
@@ -59,10 +59,10 @@ public void DropLogin(LoginProperties login)
 
     public decimal DatabaseSizeMb(string dbName)
     {
-      string query = "SELECT total_size_mb = (SUM(size) * 8. / 1024) " + 
-                     "FROM sys.master_files WITH(NOWAIT) " +
-                     $"WHERE database_id = DB_ID('{dbName}') " + 
-                     "GROUP BY database_id";
+      var query = "SELECT total_size_mb = (SUM(size) * 8. / 1024) " +
+                  "FROM sys.master_files WITH(NOWAIT) " +
+                  $"WHERE database_id = DB_ID('{dbName}') " +
```
Kaneraz, line 64: Can we parameterize dbName since it's a string value?

Kaneraz (changes requested):

src/SmoFacade/Server.cs
alexirion10: yep good point will change that

Kaneraz (approved):
Thanks, it looks good.

alexirion10:
closes https://github.com/factset/AgDatabaseMove/issues/46

PR #69, opened 2021-06-15
feat(server): add login and role

alexirion10:


Kaneraz (approved):
LGTM

Should the add role be separate? We may have multiple roles for a single login.

alexirion10:
@Kaneraz - right `addRole` could certainly be separate.  I was thinking we'd want to return the login name or properties in that case so you can do something like the following:
`public void AddRole(LoginProperties login, string role) { AssociatedLogins().Select(l => l.name = login.name).AddRole(role); }`

Kaneraz (approved):

PR #73, opened 2021-07-07
fix(backup chain): check `DatabaseBackupLsn` for log backups

AdityaNPL:
Adding a quick check that  `log_backup.DatabaseBackupLsn == fullBackup.CheckpointLsn`

Looks like I already added that logic for diff backups:
https://github.com/factset/AgDatabaseMove/blob/ecc7fe9f6ef20e3d278430add117afb3354abe05/src/BackupChain.cs#L75-L78

Unit tests are passing.

ref:
https://sqlbak.com/academy/backup-chain
image

### DIFF
--- a/src/BackupChain.cs
+++ b/src/BackupChain.cs
@@ -29,11 +29,12 @@ namespace AgDatabaseMove
         .ToList();
 
       var orderedBackups = MostRecentFullBackup(backups).ToList();
-      orderedBackups.AddRange(MostRecentDiffBackup(backups, orderedBackups.First()));
+      var fullBackup = orderedBackups.First();
+      orderedBackups.AddRange(MostRecentDiffBackup(backups, fullBackup));
 
       var prevBackup = orderedBackups.Last();
       IEnumerable<BackupMetadata> nextLogBackups;
-      while((nextLogBackups = NextLogBackup(backups, prevBackup)).Any()) {
+      while((nextLogBackups = NextLogBackup(backups, prevBackup, fullBackup)).Any()) {
         orderedBackups.AddRange(nextLogBackups);
         prevBackup = orderedBackups.Last();
       }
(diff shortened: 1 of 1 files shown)


alexirion10:
yes, technically correct... but what value does this provide?  This isn't a bug, as nothing is broken.
The backup chain is built correctly by only checking FirstLsn/LastLsn for log backups.  While also checking the full DB backup works, it's not really necessary

AdityaNPL:
For ideal cases, yes, you are correct.
However, we Identified an edge-case where users manually taking backups on single instances within an availability group can cause the LSNs to go out of sync.
(i.e. the one-off backup's First/Last LSN can be higher than the future backups' that are taken as part of the backup chain)

We have only seen this with diff backups till now, however it can happen with log backups too 
(we think it hasn't been an issue with log backups yet because we reset the backup chain with a diff backup each day at midnight)

To ensure we exclude these one-off backups which aren't part of the backup chain for restores, we want to add this bit of filtering logic.

alexirion10:
The fix for that is taking ad-hoc backups as [Copy Only backups](https://docs.microsoft.com/en-us/sql/relational-databases/backup-restore/copy-only-backups-sql-server?view=sql-server-ver15).

Any reason that can't be done?

AdityaNPL:
Ah yes, that's true and we already have a filter for that:
https://github.com/factset/AgDatabaseMove/blob/42fa22a4f59ad58ccf4b3efe2c892f436bb61963/src/SmoFacade/Database.cs#L85

Let me check and get back.

AdityaNPL:
Ah, looks like I misunderstood! 

It wasn't a manual one-off backup, it was a backup part of the normal back-up process. However, it was for the wrong server, so it shouldn't have been there.
We will need to discuss this internally as I need to share some details and I believe there might be another issue at play here.

However, in the meantime I think we can introduce this change here to make this module more stable - what are your thoughts on this?

alexirion10:
My thoughts are that we should figure out what other issues are at play first.
While this change is technically correct, it shouldn't be required

AdityaNPL:
We have identified the issue and have other changes being rolled out for the fix.

@alexirion10 in my opinion, we should merge this too as it is theoretically correct- I don't see any downsides.
What are your thoughts on this?

alexirion10:
I don't think this is needed

PR #75, opened 2021-08-09
fix(backup-chain): Add `@@ServerName` filter

AdityaNPL:
- [X] Test on stand-alone DB
- [ ] Test on AG DB

### DIFF
--- a/src/SmoFacade/Database.cs
+++ b/src/SmoFacade/Database.cs
@@ -76,13 +76,15 @@ namespace AgDatabaseMove.SmoFacade
                   "FROM msdb.dbo.backupset s " +
                   "INNER JOIN msdb.dbo.backupmediafamily m ON s.media_set_id = m.media_set_id " +
                   "WHERE s.last_lsn >= (" +
-                  "SELECT MAX(last_lsn) FROM msdb.dbo.backupset " +
-                  "WHERE [type] = 'D' " +
-                  "AND database_name = @dbName " +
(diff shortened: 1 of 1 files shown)


src/SmoFacade/Database.cs
```diff
@@ -76,13 +76,15 @@ public List<BackupMetadata> RecentBackups()
...
+                    "SELECT MAX(last_lsn) FROM msdb.dbo.backupset " +
+                    "WHERE [type] = 'D' " +
+                    "AND database_name = @dbName " +
+                    "AND is_copy_only = 0" +
+                    "AND server_name = @@ServerName " +
                   ") " +
                   "AND s.database_name = @dbName " +
-                  "AND is_copy_only = 0" +
+                  "AND is_copy_only = 0 " +
+                  "AND s.server_name = @@ServerName " +
```
alexirion10, line 87: Can you put it in all caps , that's how the t-sql is defined.
https://docs.microsoft.com/en-us/sql/t-sql/functions/servername-transact-sql?view=sql-server-ver15

alexirion10 (approved):
lgtm.  Adding @Kaneraz to the review as well

src/SmoFacade/Database.cs
AdityaNPL: Sure, done!

Kaneraz (approved):
LGTM

PR #76, opened 2021-08-10
feat(connection checks): new `CheckDBConnection()` for the `Server` and `AgDatabase` class

AdityaNPL:
This will allow us to check the DB connections (with a small connect timeout).
Helps to quickly identify issues with the connection string - like typos, wrong creds, etc.

- [X] Test with stand-alone DB
- [x] Test with AG DB

### DIFF
--- a/src/AgDatabase.cs
+++ b/src/AgDatabase.cs
@@ -233,5 +233,10 @@ namespace AgDatabaseMove
     {
       _listener.Primary.Database(Name).MultiUserMode();
     }
+
+    public void CheckDBConnections(int connectionTimeout)
+    {
+      _listener.ForEachAgInstance(server => server.CheckDBConnection(Name, connectionTimeout));
+    }
   }
 }
--- a/src/SmoFacade/Server.cs
+++ b/src/SmoFacade/Server.cs
@@ -19,10 +19,12 @@ namespace AgDatabaseMove.SmoFacade
   {
     internal readonly string _credentialName;
     internal readonly Microsoft.SqlServer.Management.Smo.Server _server;
+    private readonly string _connectionString;
 
     public Server(string connectionString, string credentialName = null)
     {
-      SqlConnection = new SqlConnection(connectionString);
+      _connectionString = connectionString;
+      SqlConnection = new SqlConnection(_connectionString);
       _server = new Microsoft.SqlServer.Management.Smo.Server(new ServerConnection(SqlConnection));
       _credentialName = credentialName;
     }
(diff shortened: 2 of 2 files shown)


src/AgDatabase.cs
```diff
@@ -233,5 +233,10 @@ public void MultiUserMode()
     {
       _listener.Primary.Database(Name).MultiUserMode();
     }
+
+    public void CheckDBConnections(int connectionTimeout)
```
lotsofspots, line 237: This is expected to throw when the first failure is encountered, right? Is that the desired behaviour, or would we want it to check all of the hosts in a cluster and give use the health of each?

AdityaNPL: That was my thinking behind it - if it can't connect to even one it should fail - i.e. raise exception

But we can refactor to parallely check each instance and return a list as a health report - could even be a list of exceptions?

AdityaNPL: For my use-case I just need the former

alexirion10 (approved):

AdityaNPL:
I'm getting some issues with AG DB testing for this change.
It looks like the connection strings in the server objects for the AG replica instances are missing the `password` parameter even though I have set it when connecting to the listener.

Looking through the code and will update once I find the reason.

AdityaNPL:
The caller needs to set `PersistSecurityInfo=True`  to be able to re-use the same connection string.
Have added this note as a comment in the code

src/SmoFacade/Server.cs
```diff
@@ -281,5 +281,23 @@ public void AddRole(LoginProperties login, RoleProperties role)
 
       matchingLogin.AddRole(role.Name);
     }
+
+    public void CheckDBConnection(string dbName, int connectionTimeout)
+    {
+      // Make sure to set `PersistSecurityInfo=True` in the connection string otherwise the password parameter might not be present
+      // https://docs.microsoft.com/en-us/dotnet/api/system.data.sqlclient.sqlconnection.connectionstring?view=dotnet-plat-ext-5.0#remarks
```
alexirion10, line 288: I'd suggest that it might be better to put this comment before the method definition, which will generate a doc comment https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/documentation-comments

This is also kinda tricky because the `PersistSecurityInfo=True` needs to be passed into the AgDatabase constructor, not method.

It might be worth saving a `SqlConnectionStringBuilder` object as a member in AgDatabase class

AdityaNPL: ah true, and then generate the connection string when needed
Let me look into it

AdityaNPL: I've added in a private member variable to store the read-only connection string - to make sure no one changes it.
It can be used to create `SqlConnectionStringBuilder`s if someone needs a modified connection sting.

Hope that's fine, let me know your thoughts

alexirion10 (approved):
lgtm

PR #77, opened 2021-08-17
feat(login): verify logins exist

slin5299:


src/Exceptions/MissingLoginException.cs
```diff
@@ -0,0 +1,13 @@
...
+  using System;
+
+
+  /// <summary>
+  ///   A base exception for the AgDatabaseMove library
+  /// </summary>
+  public class MissingLoginException : AgDatabaseMoveException
+  {
+    public MissingLoginException(string message) : base(message) { }
+  }
```
AdityaNPL, line 12: The summary might need to be changed

AdityaNPL: Same for the other exceptions too

src/SmoFacade/Listener.cs
```diff
@@ -89,6 +98,8 @@ public Listener(SqlConnectionStringBuilder connectionStringBuilder, string crede
 
     public AvailabilityGroup AvailabilityGroup { get; set; }
 
+    public IEnumerable<Server> AllInstances => _allInstances;
+
```
AdityaNPL, line 102: You might be able to do a join b/w primaries and secondaries - not 100% sure tho??

AdityaNPL (commented):
Looks good!
I just gave it a quick glance over, I'll leave it to Alex/Ryan for the final reviews.

Some quick suggestions from my side:

src/Exceptions/MissingLoginException.cs
slin5299: yup, done.

src/SmoFacade/Listener.cs
slin5299: I removed this since it is no longer needed.

alexirion10: Yeah I'm not sure we need an additional variable for this.  You should be able to join the primary and secondary members inline as a property

src/SmoFacade/Listener.cs
```diff
@@ -74,10 +74,12 @@ public Listener(SqlConnectionStringBuilder connectionStringBuilder, string crede
...
-      foreach(var secondaryName in secondaryNames)
-        _secondaries.Add(AgInstanceNameToServer(ref connectionStringBuilder,
-                                                secondaryName,
-                                                credentialName));
+      foreach(var secondaryName in secondaryNames) {
+        var secondary = AgInstanceNameToServer(ref connectionStringBuilder,
+                                               secondaryName,
+                                               credentialName);
+        _secondaries.Add(secondary);
+      }
```
alexirion10, line 82: can you revert this back :)

alexirion10 (commented):
Looks pretty good to me!  Have you had a chance to test this on any clusters yet?

src/SmoFacade/Listener.cs
slin5299: done

src/AgDatabase.cs
```diff
@@ -233,5 +237,48 @@ public void MultiUserMode()
...
+        }
+      });
+
+      if(exceptions.Count > 0) throw new AggregateException(exceptions);
+    }
+
+    public byte[] GetSid(string dbName)
+    {
+      return _listener.Primary.Logins.Single(d => d.Name == dbName).Sid;
+    }
```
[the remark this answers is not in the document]
slin5299: done

src/AgDatabase.cs
```diff
@@ -12,6 +12,8 @@ namespace AgDatabaseMove
   using System.Data.SqlClient;
   using System.Linq;
   using System.Threading;
+  using Exceptions;
+  using Microsoft.SqlServer.Management.SqlParser.SqlCodeDom;
```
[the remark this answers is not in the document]
slin5299: I don't remember, it's been removed

slin5299:
It would be hard to test on a cluster since AgDatabase doesn't have access to admin credentials and Catalogue. I've been hard coding individual listeners and databases to run tests.

alexirion10:
@slin5299 - yep exactly, that's good!  Just wanted to make sure it was tested on a real cluster

src/AgDatabase.cs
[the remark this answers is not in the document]
alexirion10: So I actually realized we already have a function that returns `LoginProperty`... actually a list of all login-properties, but they can still be filtered out in the same way. Right?
So I don't think we need this at all!  my b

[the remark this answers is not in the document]
slin5299: Yep, `AssociatedLogins()` returns an IEnumerable, which can be filtered. I'll remove this function.

alexirion10 (approved):

PR #78, opened 2021-08-23
feat(logins): make sids optional

slin5299:
Currently, a sid is required to create a new login. This change makes it possible to create logins without providing a sid. If there is no sid provided, a new one will be generated.

### DIFF
--- a/src/SmoFacade/Login.cs
+++ b/src/SmoFacade/Login.cs
@@ -104,11 +104,15 @@ namespace AgDatabaseMove.SmoFacade
     {
       var login = new Microsoft.SqlServer.Management.Smo.Login(server._server, loginProperties.Name) {
         LoginType = loginProperties.LoginType,
-        Sid = loginProperties.Sid,
         DefaultDatabase = loginProperties.DefaultDatabase
       };
 
-      if(loginProperties.LoginType == LoginType.SqlLogin) {
+      if (loginProperties.Sid != null)
+      {
+        login.Sid = loginProperties.Sid;
+      }
+
+      if (loginProperties.LoginType == LoginType.SqlLogin) {
         if(loginProperties.PasswordHash != null)
           login.Create(loginProperties.PasswordHash, LoginCreateOptions.IsHashed);
         else if(loginProperties.Password != null)


src/SmoFacade/Login.cs
```diff
@@ -104,11 +104,15 @@ public void Drop()
...
         LoginType = loginProperties.LoginType,
-        Sid = loginProperties.Sid,
         DefaultDatabase = loginProperties.DefaultDatabase
       };
 
-      if(loginProperties.LoginType == LoginType.SqlLogin) {
+      if (loginProperties.Sid != null)
+      {
+        login.Sid = loginProperties.Sid;
+      }
```
AdityaNPL, line 113: Can't we just keep it as it is?
It will set it to `null` if it is not present - or am I missing something here?

slin5299: In the current code, if you try to create a login without a sid (or without a DefaultDatabase or LoginType), it will throw a null value exception.

AdityaNPL: ah, so you can't set `Sid` to `null` but you can leave it uninitialised?

AdityaNPL: Which makes sense as the default looks like it should be a `byte[]`
https://docs.microsoft.com/en-us/dotnet/api/microsoft.sqlserver.management.smo.login.sid?view=sql-smo-160#Microsoft_SqlServer_Management_Smo_Login_Sid

Can we do this instead?

```csharp
Sid = loginProperties.Sid ?? byte[]
```

(Incase you have not come across `??` yet: https://stackoverflow.com/q/446835)

slin5299: You can't set `Sid` to `null` and you also can't leave it uninitialized. Both will result in a null exception being thrown. I did some tests with your suggestion, but I ran into some issues. If I set it to an empty array, it throws another exception saying "Supplied parameter should be binary(16)". If I set it to `btye[16]`, this creates a sid with all zeros and it throws an exception saying "supplied parameter sid is already in use".

With this PR, if the Sid is uninitialized, it will automatically create a new one.

PR #80, opened 2021-09-22
fix(server): windows path on linux runtime

alexirion10:


src/SmoFacade/Server.cs
```diff
@@ -130,24 +142,24 @@ private DefaultFileLocations DefaultFileLocations()
...
+      string result = null;
+      if(!string.IsNullOrWhiteSpace(backupDirectoryQuery)) {
+        using var cmd = SqlConnection.CreateCommand();
+        cmd.CommandText = backupDirectoryQuery;
+        using var reader = cmd.ExecuteReader();
+        if(reader.Read())
+          result = (string)reader[0];
+      }
+
+      result ??= _server.BackupDirectory;
```
AdityaNPL, line 156: Oooo, I didn't know you could do this - `??=`

src/SmoFacade/Server.cs
```diff
@@ -15,81 +15,93 @@ namespace AgDatabaseMove.SmoFacade
...
-                  "AND SUBSTRING(CONVERT(VARCHAR, SERVERPROPERTY('InstanceDefaultDataPath')), 1, 3) = dovs.volume_mount_point"; // Disk Drive equals
-
-      using var cmd = SqlConnection.CreateCommand();
-      cmd.CommandText = query;
-      using var reader = cmd.ExecuteReader();
-      reader.Read();
-      return (int)reader["free_size_mb"];
+      if(Path.IsPathRooted(physicalName))
+        return Path.GetFileName(physicalName) ??
+               throw new InvalidBackupException($"Physical name in backup is incomplete: {physicalName}");
```
AdityaNPL, line 99: What happens if it is not rooted?
Should we throw an exception?

src/SmoFacade/Server.cs
alexirion10: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-coalescing-operator

src/SmoFacade/Server.cs
alexirion10: See code lines below.... chop off everything before `\\` to get filename

AdityaNPL: Right for the windows paths, but how bout linux paths?

alexirion10: I'm pretty sure `IsPathRooted` == true even on windows, but I need to test that more thoroughly

PR #81, opened 2021-09-22
fix(rollback): servername filter on query

alexirion10:
if you use `@@servername` in the filter, then the backupChain is not retrieved correctly for databases restored from another server


AdityaNPL (commented):
LGTM!

PR #83, opened 2021-11-15
fix(database): drop retry loop SyncState

alexirion10:


src/SmoFacade/Database.cs
```diff
@@ -45,21 +45,30 @@ public void RestoreWithRecovery()
...
-      policy.Execute(() => {
-        if(string.IsNullOrEmpty(_database.AvailabilityGroupName))
-          return;
-        if(_database.AvailabilityDatabaseSynchronizationState >
-           AvailabilityDatabaseSynchronizationState.NotSynchronizing)
-          throw new
-            TimeoutException($"Cannot kill the database {Name} until it has been removed from the AvailabilityGroup.");
+      policyPrep.Execute(() => {
+        if (!string.IsNullOrEmpty(_database.AvailabilityGroupName))
+          throw new Exception($"Cannot kill the database {Name} until it has been removed from the AvailabilityGroup");
```
Kaneraz, line 55: Can we throw something more specific than exception? Or do we truly want to retry on every type of exception?

alexirion10: What other exception types are possible here?  All its doing is retrieving an Enum property.
I'm not sure what benefit a different exception type would make here

Kaneraz: That's the point, I don't care to dig into the database object and likely into SMO to find out what other exceptions we may be retrying. Using a uniquely typed exception for what we want to retry on makes more sense, unless we're saying we want to retry on everything even not knowing what everything is.

alexirion10: I want this to retry on everything, yes.
If SMO fails to retrieve one of these properties, throwing anything other than `PropertyCannotBeRetrievedException`, it should retry

Kaneraz (approved):

PR #86, opened 2021-11-16
feat(agdb): node status

alexirion10:
This is probably a little lazy, but a possible method to determine whether all AvailabilityGroup node are online prior to a restore or not


Kaneraz (commented):
In another case I use a concurrent bag to gather the results. It'd be better to use that than relying on exception handling. 

https://github.com/factset/AgDatabaseMove/blob/42a5e28379d6ee461d2fa4bc1353ed8a62540cb6/src/AgDatabase.cs#L125

PR #88, opened 2021-12-09
fix(backup-chain): checking checkpoint LSNs

AdityaNPL:
Making sure DB-LSNs point to last Full backup's Checkpoint LSN

---

Testing
- [X] Unit Tests Pass
- [x] Local MSSQL DB

### DIFF
--- a/src/BackupChain.cs
+++ b/src/BackupChain.cs
@@ -29,11 +29,12 @@ namespace AgDatabaseMove
         .ToList();
 
       var orderedBackups = MostRecentFullBackup(backups).ToList();
-      orderedBackups.AddRange(MostRecentDiffBackup(backups, orderedBackups.First()));
+      var fullBackup = orderedBackups.First();
+      orderedBackups.AddRange(MostRecentDiffBackup(backups, fullBackup));
 
       var prevBackup = orderedBackups.Last();
       IEnumerable<BackupMetadata> nextLogBackups;
-      while((nextLogBackups = NextLogBackup(backups, prevBackup)).Any()) {
+      while((nextLogBackups = NextLogBackup(backups, prevBackup, fullBackup)).Any()) {
         orderedBackups.AddRange(nextLogBackups);
         prevBackup = orderedBackups.Last();
       }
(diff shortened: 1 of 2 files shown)


src/BackupChain.cs
```diff
@@ -84,11 +85,12 @@ private static IEnumerable<BackupMetadata> MostRecentFullBackup(IEnumerable<Back
...
     private static IEnumerable<BackupMetadata> NextLogBackup(IEnumerable<BackupMetadata> backups,
-      BackupMetadata prevBackup)
+      BackupMetadata prevBackup, BackupMetadata lastFullBackup)
     {
       // also gets all the stripes of the next backup
-      return backups.Where(b => b.BackupType == BackupFileTools.BackupType.Log &&
-                                prevBackup.LastLsn >= b.FirstLsn && prevBackup.LastLsn + 1 < b.LastLsn);
+      return backups.Where(b => b.BackupType == BackupFileTools.BackupType.Log 
+                                && prevBackup.LastLsn >= b.FirstLsn && prevBackup.LastLsn + 1 < b.LastLsn
+                                && b.DatabaseBackupLsn == lastFullBackup.CheckpointLsn);
```
alexirion10, line 93: While I think this is valid - I'm not seeing why it's needed.  
In what scenario does `b.DatabaseBackupLsn != lastFullBackup.CheckpointLsn` via the results of the SQL query to retrieve SQL backup data?

AdityaNPL: (Sorry for the late reply on this)

I'm not sure I fully understand your question - can you please elaborate?

We need this filter because:
When a database has been restored onto the same host, we have seen that sometimes there are some backups that aren't part of the chain. To filter these out we can use this check. 
(Note: these odd backups are also not copy-only)

We have this filter here and not in the SQL query because:
The [SQL query](https://github.com/factset/AgDatabaseMove/blob/fix/backup-chain/checkpoint-lsn-filter/src/SmoFacade/Database.cs#L82) is executed on an individual instance - we don't want to filter the backups there as that instance's full-backup might not be the full backup we want to use for the backup chain (i.e. not the correct full-backup)
Therefore, using an incorrect fullbackup's checkpoint LSN might result in required backups being filtered out.

Kaneraz: I've been thinking about this. It may be time to bite the bullet and throw out my query and design. 

Instead of trying to be fancy and retrieve all the backups in one parallel go, we can build the chain in multiple queries. Then instead of going from the last full backup, we just get the current [LSN properties](https://docs.microsoft.com/en-us/sql/relational-databases/system-dynamic-management-views/sys-dm-db-log-stats-transact-sql?view=sql-server-ver15) and query the backups based on the [database backup LSN and checkpoint LSN](https://www.mssqltips.com/sqlservertip/3209/understanding-sql-server-log-sequence-numbers-for-backups/). This current back plan should resolve the other issues we've seen.

alexirion10: That seems reasonable as well.  Query performance would take a slight hit, but it has the potential to build a more reliable LSN backup chain

AdityaNPL: Agreed.
I'll keep it on our backlog for now and come back to this when I get some time

PR #89, opened 2021-12-14
feat(db): drop retry logic

alexirion10:
if `restoring` DB is good to drop
if `catch(SmoException e) when(e.Message.Contains("Cannot access properties or methods"))` DB is good to drop

### DIFF
--- a/src/SmoFacade/Database.cs
+++ b/src/SmoFacade/Database.cs
@@ -45,24 +45,7 @@ namespace AgDatabaseMove.SmoFacade
     /// </summary>
     public void Drop()
     {
-      var policyPrep = Policy
-        .Handle<Exception>()
-        .WaitAndRetry(5, retryAttempt => TimeSpan.FromMilliseconds(Math.Pow(10, retryAttempt)));
-
-      // ensure database is not in AvailabilityGroup, WaitAndRetry loop for each instance to sync
-      policyPrep.Execute(() => {
-        try {
-          if(!string.IsNullOrEmpty(_database.AvailabilityGroupName))
-            throw
-              new Exception($"Cannot kill the database {Name} until it has been removed from the AvailabilityGroup");
-          if(_database.AvailabilityDatabaseSynchronizationState >
-             AvailabilityDatabaseSynchronizationState.NotSynchronizing)
-            throw new
-              Exception($"Cannot kill the database {Name} until AvailabilityDatabaseSynchronizationState is inaccessible");
-        }
-        catch(
-          PropertyCannotBeRetrievedException) { } // good here, AvailabilityDatabaseSynchronizationState unavailable means it has no state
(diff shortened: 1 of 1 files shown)


src/SmoFacade/Database.cs
```diff
@@ -134,5 +117,31 @@ public void MultiUserMode()
       _database.DatabaseOptions.UserAccess = DatabaseUserAccess.Multiple;
       _database.Alter(TerminationClause.RollbackTransactionsImmediately);
     }
+
+    private void ThrowIfUnreadyToDrop()
+    {
+      if(Restoring)
+        return; // restoring state means we're good to drop
```
AdityaNPL, line 124: If it's restoring, shouldn't we wait until it is finished?

src/SmoFacade/Database.cs
```diff
@@ -134,5 +117,31 @@ public void MultiUserMode()
...
+            throw
+              new Exception($"Cannot kill the database {Name} until it has been removed from the AvailabilityGroup");
+          if(_database.AvailabilityDatabaseSynchronizationState >
+             AvailabilityDatabaseSynchronizationState.NotSynchronizing)
+            throw new
+              Exception($"Cannot kill the database {Name} until AvailabilityDatabaseSynchronizationState is inaccessible");
+        }
+        catch(
+          PropertyCannotBeRetrievedException) { } // good here, AvailabilityDatabaseSynchronizationState unavailable means it has no state
+        catch(SmoException e) when(e.Message.Contains("Cannot access properties or methods")) { }
```
AdityaNPL, line 143: Should we log the error message in these cases?
I just feel catching and ignoring exceptions may lead to headaches down the line when trying to debug prod issues

AdityaNPL (commented):
Have a few questions

src/SmoFacade/Database.cs
alexirion10: no - restoring state doesn't necessarily mean that it's currently writing a backup.  
Restoring state happens on secondary nodes when the DB is removed from the availability group. 

See:  https://docs.microsoft.com/en-us/sql/database-engine/availability-groups/windows/remove-a-primary-database-from-an-availability-group-sql-server?view=sql-server-ver15#FollowUp

src/SmoFacade/Database.cs
alexirion10: I see no need to log here, these are cases where we can safely proceed

AdityaNPL (approved):
LGTM!

alexirion10:
@Kaneraz - could you take a look at this?

Kaneraz (commented):
Oops, late to hit submit here. The comments we discussed.

Kaneraz (approved):
Thanks, looks good

PR #94, opened 2022-03-21
fix(agdatabase): drop sync state

alexirion10:


src/AgDatabase.cs
```diff
@@ -156,7 +153,7 @@ public void DropLogin(LoginProperties login)
 
     public void DropAllLogins()
     {
-      foreach(var loginProp in AssociatedLogins()) DropLogin(loginProp);
```
Kaneraz, line 159: Originally we were dropping logins by reading the logins on the primary database which should be accessible while the others aren't. At this point, with the databases being removed from the AG are databases on all nodes readable?

Kaneraz (commented):
A couple notes of interest, but looks good. Just moving around the checks we expect to be doing and making use of ?. incase something goes wrong.

src/AgDatabase.cs
alexirion10: Well, the idea here is to drop the logins before the database, so they are all readable.
What you mentioned, `reading the logins on the primary database which should be accessible while the others aren't` is true.  But the logic in the code was only deleting those logins on the Primary node, not all nodes.

The other case we have in our application is... we can delete the DB, then do a 
agDb.DropLogin("<login_name_here>");
agDb.DropLogin("<login_name_here>_reader");

which does not require reading secondary database properties.  Its a no-throw attempt at removing Server level logins only

src/AgDatabase.cs
```diff
@@ -89,9 +89,6 @@ public bool Exists()
     /// </summary>
     public void Delete()
     {
-      // Deleting the database while it is initializing will leave it in a state where system redo threads are stuck.
-      // This leaves the database in a state that a SQL Server service restart prior to deletion.
-      _listener.ForEachAgInstance(WaitForInitialization);
```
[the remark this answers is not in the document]
alexirion10: Can you clarify `Do we need to pull this method now`?  Yes we could delete `WaitForInitialization` & `IsInitializing` if we wanted.
It's covered in the `Database` class, so not really needed in the AgDatabase IMO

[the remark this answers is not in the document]
Kaneraz: Ya, that's what I was thinking. May as well clear them out if they're private and not part of the interface exposed to the consumers.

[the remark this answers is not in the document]
alexirion10: removed unused code in https://github.com/factset/AgDatabaseMove/pull/94/commits/e6d0657d3e9ceb5085067dde967f891b2a667020

src/SmoFacade/AvailabilityGroup.cs
```diff
@@ -64,9 +64,19 @@ public bool IsInitializing(string dbName)
...
 
-      return availabilityDatabase.SynchronizationState == MsftSmo.AvailabilityDatabaseSynchronizationState.Initializing;
+      try {
+        return availabilityDatabase.SynchronizationState ==
+               MsftSmo.AvailabilityDatabaseSynchronizationState.Initializing;
+      }
+      catch(
+        MsftSmo.PropertyCannotBeRetrievedException) { } // good here, AvailabilityDatabaseSynchronizationState unavailable means it has no state
+      catch(MsftSmo.SmoException e) when(e.Message.Contains("Cannot access properties or methods")) { }
+      catch(Exception) { return true; }
```
AdityaNPL, line 77: Just curious, why are we returning `true` when there is an error?

src/SmoFacade/Database.cs
```diff
@@ -40,6 +40,17 @@ public void RestoreWithRecovery()
       _database.Refresh();
     }
 
+    public void DropAssociatedLogins()
+    {
+      var logins = Users.Where(u => u.Login != null && u.Login.Name != "sa")
```
AdityaNPL, line 45: Small nit pick, can we pls place `"sa"` in a variable called `SystemAdminAccountName` or something like that?
Just feel it would be easier to understand what it is for engineers not familiar with SQL Server logins.

I had to google what "sa" was when reading through this 😅

Kaneraz (approved):

PR #95, opened 2022-03-28
feat(polly): retry db drop wait longer

alexirion10:
4 retries starting at 10ms:
10ms, 100ms, 1000ms (1 second) 

6 retries starting at 5ms:
5ms, 25ms, 125ms, 625ms, 3125ms (3.1 seconds), 15625ms (15.625 seconds)


Kaneraz (approved):
Sounds good.

PR #96, opened 2022-04-01
feat(login): create new sql login

mja9:
Added functionality to create a new Sql login replicated on all replicas of an availability group. Requires creation of login on primary, saving SID, then using the saved SID to create the same login on secondary's.
Currently extends `AddLogin` to detect a case where caller may be trying to create a Sql login which does not yet exist. 

Added integration test for `AgDatabase` to test functionality described above. Tests for existence of logins, logins are created on more than just the primary, and verifies Sid of logins match.

These changes pass the tests included. I also manually verified creation of logins with matching Sid on each replica of availability group.

### DIFF
--- a/tests/AgDatabaseMove.Integration/Fixtures/TestAgDatabaseFixture.cs
+++ b/tests/AgDatabaseMove.Integration/Fixtures/TestAgDatabaseFixture.cs
@@ -1,0 +1,48 @@
+﻿using AgDatabaseMove.Integration.Config;
+using System;
+
+namespace AgDatabaseMove.Integration.Fixtures
+{
+  public class TestAgDatabaseFixture : TestConfiguration<TestAgDatabaseConfig>, IDisposable
+  {
+    public TestConfiguration<TestLoginConfig> _login;
+
+    public AgDatabase _agDatabase;
+
+    public TestAgDatabaseFixture() : base("TestAgDatabase")
+    {
+      _agDatabase = ConstructAgDatabase();
+      _login = new TestConfiguration<TestLoginConfig>("TestLogin");
+    }
+
+    // Public fields for testing new sql login creation.
+    public string _loginPassword => _login._config.Password;
+
+    public string _loginName => _login._config.LoginName;
+
+    public string _loginDefaultDatabase => _login._config.DefaultDatabase;
+
+    public AgDatabase ConstructAgDatabase()
+    {
+      var dbConfig = new DatabaseConfig
+      {
+        BackupPathSqlQuery = _config.BackUpPathSqlQuery,
+        ConnectionString = _config.ConnectionString,
+        DatabaseName = _config.DatabaseName,
+        CredentialName = _config.CredentialName
+      };
+
+      return new AgDatabase(dbConfig);
+    }
+    public void Dispose()
+    {
+      // Cleanup new Sql logins created.
+      _agDatabase._listener.Primary._server.Logins[_loginName]?.DropIfExists();
+      foreach(var server in _agDatabase._listener.Secondaries) {
+        server._server.Logins[_loginName]?.DropIfExists();
+      }
+
+      _agDatabase.Dispose();
+    }
+  }
+}
--- a/tests/AgDatabaseMove.Integration/Fixtures/TestAgDatabaseFixture.cs
+++ b/tests/AgDatabaseMove.Integration/Fixtures/TestAgDatabaseFixture.cs
@@ -55,8 +55,8 @@
 
     public void Dispose()
     {
-      _agDatabase.DropLogin(new LoginProperties { Name = _loginConfig.LoginName });
-      _agDatabase.Dispose();
+      _agDatabase?.DropLogin(new LoginProperties { Name = _loginConfig.LoginName });
+      _agDatabase?.Dispose();
     }
   }
 }
--- a/tests/AgDatabaseMove.Integration/Fixtures/TestAgDatabaseFixture.cs
+++ b/tests/AgDatabaseMove.Integration/Fixtures/TestAgDatabaseFixture.cs
@@ -1,47 +1,61 @@
-﻿using AgDatabaseMove.Integration.Config;
-using System;
-
-namespace AgDatabaseMove.Integration.Fixtures
+﻿namespace AgDatabaseMove.Integration.Fixtures
 {
-  public class TestAgDatabaseFixture : TestConfiguration<TestAgDatabaseConfig>, IDisposable
+  using Config;
+  using System;
+  using System.Collections.Generic;
+  using Microsoft.SqlServer.Management.Smo;
+  using SmoFacade;
+  using System.Linq;
+
+  public class TestAgDatabaseFixture : IDisposable
   {
-    public TestConfiguration<TestLoginConfig> _login;
+    public readonly TestAgDatabaseConfig _agConfig =
+      new TestConfiguration<TestAgDatabaseConfig>("TestAgDatabase")._config;
+
+    public readonly TestLoginConfig _loginConfig =
+      new TestConfiguration<TestLoginConfig>("TestLogin")._config;
 
     public AgDatabase _agDatabase;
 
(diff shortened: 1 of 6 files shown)


alexirion10 (approved):
lgtm!  Have you tested?

mja9:
> lgtm! Have you tested?

Yes, I tested with a quick main prog. I was able to verify the login existed on all replicas, the SID's matched, and created a dummy user to verify I could login on the primary. Looking to add an integration test here before merging this in.

tests/AgDatabaseMove.Integration/TestAgDatabase.cs
```diff
@@ -0,0 +1,75 @@
...
+      _testAgDatabaseFixture = testAgDatabaseFixture;
+    }
+
+    private AgDatabase AgDatabase => _testAgDatabaseFixture._agDatabase;
+
+    private string LoginName => _testAgDatabaseFixture._loginName;
+
+    private string LoginPassword => _testAgDatabaseFixture._loginPassword;
+
+    private string DefaultDatabase => _testAgDatabaseFixture._loginDefaultDatabase;
```
alexirion10, line 24: What benefit does this provide?  All of these values can be accessed from within the class already

tests/AgDatabaseMove.Integration/TestAgDatabase.cs
```diff
@@ -0,0 +1,75 @@
...
+
+    private void CreateNewSqlLogin(LoginProperties login)
+    {
+      AgDatabase.AddLogin(login);
+    }
```
alexirion10, line 29: Seems like this can be inlined...? Why do you need a separate function?

tests/AgDatabaseMove.Integration/TestAgDatabase.cs
```diff
@@ -0,0 +1,75 @@
...
+
+    private void TestLoginExistsOnAg(List<Smo.Login> createdLogins)
+    {
+      createdLogins.ForEach(Assert.NotNull);
+    }
```
alexirion10, line 34: same as above.  Why do we need a separate function here?

tests/AgDatabaseMove.Integration/TestAgDatabase.cs
```diff
@@ -0,0 +1,75 @@
...
+        Name = LoginName,
+        Password = LoginPassword,
+        LoginType = Smo.LoginType.SqlLogin,
+        DefaultDatabase = DefaultDatabase
+      };
+
+      CreateNewSqlLogin(newSqlLogin);
+      var createdLogins = GetCreatedLogins();
+      TestLoginExistsOnAg(createdLogins);
+      TestLoginSidsMatch(createdLogins);
```
alexirion10, line 72: Ideally, a test will only test 1 thing.  You are testing 3 or 4 different parts of the login creation/exists/sids here.  Can you break this up into several different tests, each with a single responsibility?

tests/AgDatabaseMove.Integration/TestAgDatabase.cs
```diff
@@ -0,0 +1,75 @@
...
+      Assert.True(logins.Count > 1);
+      return logins;
+    }
+
+    private void TestLoginSidsMatch(List<Smo.Login> createdLogins)
+    {
+      for (int i = 0; i < createdLogins.Count - 1; i++)
+      {
+        Assert.Equal(createdLogins[i].Sid, createdLogins[i + 1].Sid);
+      }
```
alexirion10, line 55: mmm gross.... `Count - 1` && comparing `i == i+1`.
XUnit has special Assert overloads for collections.  Something like this:
`Assert.All(list, item =>  item.Sid.Equals(expected));`

tests/AgDatabaseMove.Integration/TestAgDatabase.cs
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    private List<Smo.Login> GetCreatedLogins()
+    {
+      List<Smo.Login> logins = new List<Smo.Login>();
+      logins.Add(AgDatabase._listener.Primary._server.Logins[LoginName]);
+      foreach(var server in AgDatabase._listener.Secondaries)
+      {
+        logins.Add(server._server.Logins[LoginName]);
+      }
```
alexirion10, line 43: The purpose of AgDatabase is to abstract the Primary vs N number of secondary nodes.
The test here would be better if it didn't take advantage of `internal` properties.  Perhaps `_listener` and `_server` should be private members

alexirion10 (changes requested):

tests/AgDatabaseMove.Integration/TestAgDatabase.cs
mja9: I believe `_listener` is used in another test class. It could be made `private` instead, but that should be done in another PR. `_server` on the other hand has multiple use-cases requiring that it stays at the same visibility.

Kaneraz (approved):
Thanks for the added tests!

tests/AgDatabaseMove.Integration/Fixtures/TestAgDatabaseFixture.cs
```diff
@@ -0,0 +1,62 @@
...
+        logins.Add(server.Logins
+                     .SingleOrDefault(l => l.Name.Equals(_loginConfig.LoginName, StringComparison.InvariantCultureIgnoreCase)));
+      }
+
+      return logins;
+    }
+
+    public void Dispose()
+    {
+      _agDatabase.DropLogin(new LoginProperties { Name = _loginConfig.LoginName });
```
AdityaNPL, line 58: What happens if this line throws an exception when `Dispose()` is called from the destructor?

I guess in this context, it's not that big of an issue as it's just a test instance.
But I think it's something worth highlighting and something to be aware of.

https://stackoverflow.com/questions/1030455/how-to-handle-exception-thrown-from-dispose

In this instance, maybe we can clean up the logins explicitly instead after each test?

AdityaNPL (commented):
changes look good, just a small thing in the tests

tests/AgDatabaseMove.Integration/Fixtures/TestAgDatabaseFixture.cs
alexirion10: yeah probably better to do this explicitly, or make it no-throw

tests/AgDatabaseMove.Integration/TestAgDatabase.cs
```diff
@@ -0,0 +1,30 @@
...
+    {
+      _fixture = testAgDatabaseFixture;
+    }
+
+    [Fact]
+    public void TestLoginsExist()
+    {
+      _fixture._agDatabase.ContainsLogin(_fixture._loginConfig.LoginName);
+      Assert.True(true);
+    }
```
alexirion10, line 21: `Assert.True(true)` ?  Can you assert ContainsLogin == true?

mja9: `ContainsLogin` doesn't return anything, it just throws if there's an error https://github.com/factset/AgDatabaseMove/blob/ad13f56971f5fa176ea523c0206de0a43e9a99e8/src/AgDatabase.cs#L245.

I don't think the `Assert` is necessary for the test to pass though, so I can just remove it.

tests/AgDatabaseMove.Integration/Fixtures/TestAgDatabaseFixture.cs
mja9: https://github.com/factset/AgDatabaseMove/blob/ad13f56971f5fa176ea523c0206de0a43e9a99e8/src/SmoFacade/Server.cs#L59-L62

I'm not too concerned about dropping throwing an exception here. The logic above should fail gracefully in the event the login doesn't exist. Maybe the database connection could give out before the everything gets disposed properly?

AdityaNPL (approved):
:shipit:

PR #100, opened 2022-04-21
feat(AgDb): AddUser

mja9:
Changes include functionality to add user with specific database roles/permissions.
User creation modelled after login creation: https://github.com/factset/AgDatabaseMove/blob/b9191b660e84acb691c0e82d21bd4064845563a7/src/SmoFacade/Login.cs#L37-L41

Testing:
Tested with the test cases written below as well as a separate main prog which uses the created user/login to connect to the database and performs read-only and execute operations.


tests/AgDatabaseMove.Integration/TestAgDatabase.cs
```diff
@@ -25,5 +25,28 @@ public void TestLoginSidsMatch()
...
+      Assert.True(user.EnumRoles().Contains("db_datareader"));
+    }
+
+    [Fact]
+    public void TestUserHasPermissions()
+    {
+      var db = _fixture._agDatabase._listener.Primary.Database(_fixture._loginConfig.DefaultDatabase)._database;
+      var permissions = db.EnumDatabasePermissions(_fixture._loginConfig.LoginName);
+      Assert.NotNull(permissions.FirstOrDefault(p => p.PermissionType.Execute));
+    }
```
mja9, line 50: These tests use `internal` fields. Accessing these `smo `objects could be moved to `TestAgDatabaseFixture`, or I could add accessors for roles and permissions to the `User` class instead. 

Either way I'll need to get `_agDatabase._listener.Primary` since user creation will be confined to the primary instance.

src/SmoFacade/Database.cs
```diff
@@ -50,6 +50,20 @@ public void DropAssociatedLogins()
         });
     }
 
+    public void AddUser(UserProperties user)
+    {
+      var matchingUser =
+        Users.SingleOrDefault(u => u.Name.Equals(user.Name, StringComparison.InvariantCultureIgnoreCase)) ??
+        new User(user, _server, this);
+    }
```
AdityaNPL, line 58: You could return back the user - will allow the caller to run consecutive functions?

`db.AddUser("...").DoSomething()`

src/SmoFacade/User.cs
```diff
@@ -1,18 +1,62 @@
...
+  using System.Collections.Generic;
+  using Smo = Microsoft.SqlServer.Management.Smo;
+
+  public class UserProperties
+  {
+    public string Name { get; set; }
+    public IEnumerable<RoleProperties> Roles { get; set; }
+    public Smo.DatabasePermissionSet Permissions { get; set; }
+    public string LoginName { get; set; }
+  }
```
AdityaNPL, line 12: what's the diff between `LoginName` and `Name`?

src/AgDatabase.cs
```diff
@@ -184,6 +184,16 @@ private void AddNewSqlLogin(LoginProperties login)
...
 
+    public void AddUser(UserProperties user)
+    {
+      _listener.Primary.AddUser(user, Name);
+    }
+
+    public void DropUser(UserProperties user)
+    {
+      _listener.Primary.DropUser(user, Name);
+    }
```
AdityaNPL, line 195: Shouldn't we propagate this to all servers, not just primary?

AdityaNPL (commented):
Couple of suggestions/questions

src/SmoFacade/User.cs
mja9: `LoginName` is referring to the name of the associated login, `Name` refers to the name of the user.

src/AgDatabase.cs
alexirion10: Users are a database level property which synchronize through the AG
https://www.sqlshack.com/synchronize-logins-between-availability-replicas-in-sql-server-always-on-availability-group/

mja9: This was designed to be called once the database has already joined the availability group, at which point the database-level object will  be replicated on the secondary's. I haven't tested an implementation where the database is not joined to the ag, but I assume in that case propagating this to all servers would be appropriate.

The current implementation could be extended to handle this case as well. It would need to detect whether or not the database is an availability database.

src/SmoFacade/User.cs
```diff
@@ -1,18 +1,62 @@
...
+
+    public void GrantPermission(Smo.DatabasePermissionSet permissions)
+    {
+      _database._database.Grant(permissions, Name);
+    }
+
+    public void Drop()
+    {
+      _user.Drop();
+    }
```
alexirion10, line 58: I'm not sure what this does... Does calling `user.drop` remove it from the database? (ie. does this user object have a server/database property tied to it?)

src/SmoFacade/User.cs
```diff
@@ -1,18 +1,62 @@
...
       _server = server;
+      _database = database;
+    }
+
+    public User(UserProperties userProperties, Server server, Database database)
+    {
+      _server = server;
+      _database = database;
+      ConstructUser(userProperties);
     }
```
alexirion10, line 33: I see what you're doing here...  I have a suggestion, which may or may not be better, but here's what I'm thinking:

Leave the `User` class as a POD (plain old data type).  The User creation code can live in a private/internal method on the `Database` class, where this is currently being called from.

Mainly, I think it's too easy to misuse the class constructor here unintentionally creating a new user

src/SmoFacade/Server.cs
```diff
@@ -295,6 +295,29 @@ public void AddRole(LoginProperties login, RoleProperties role)
       matchingLogin.AddRole(role.Name);
     }
 
+    public User AddUser(UserProperties user, string database)
+    {
+      var db = Databases.SingleOrDefault(d => d.Name.Equals(database, StringComparison.InvariantCultureIgnoreCase));
```
alexirion10, line 300: Can this re-use code here https://github.com/factset/AgDatabaseMove/blob/master/src/SmoFacade/Server.cs#L163 ?

src/SmoFacade/Server.cs
```diff
@@ -295,6 +295,29 @@ public void AddRole(LoginProperties login, RoleProperties role)
...
 
+    public User AddUser(UserProperties user, string database)
+    {
+      var db = Databases.SingleOrDefault(d => d.Name.Equals(database, StringComparison.InvariantCultureIgnoreCase));
+      if (db == null)
+      {
+        throw new Exception("No matching database found");
+      }
+
+      return db.AddUser(user);
```
slin5299, line 306: This is fine but it could be more concise

`db == null ? throw new Exception("No matching database found"): return db.AddUser(user)`

Same for `DropUser`

alexirion10: I think you got the wrong variable here.  `null_ ? ... : ...` is always going to be null

src/SmoFacade/User.cs
mja9: From my understanding, the user object has a database property tied to it which has a server property tied to it, yes.
I believe the drop is performed specifically from a non-query statement execution, i.e. `DROP USER _username`_

src/SmoFacade/User.cs
mja9: I'm inclined to agree. I was modeling after the login creation here, but I think constructor logic should be mostly confined to actions intended to be performed on every instantiation of that object.

src/SmoFacade/Server.cs
mja9: Yes it can!

src/SmoFacade/Server.cs
mja9: I didn't think you could optionally return or throw with a ternary operator but apparently you can! 
https://stackoverflow.com/questions/42209135/c-sharp-operator-and-throw-exception
The same won't work for `DropUser` since it's `void`.

On further review though, I think I'll remove these methods from `Server` outright, since that seems to be more aligned with the overall design in the repo. It makes more sense either way since users are strictly database objects. 

Additionally, the method `Database()` in server https://github.com/factset/AgDatabaseMove/blob/master/src/SmoFacade/Server.cs#L163
simply returns null in the case a database can't be found. Checking if the database exists relies on this, but the other methods either mark the result of `Database()` as nullable or blindly call a method on it. 

Perhaps `Database()` should explicitly throw? Certain methods like `DropAllLogins` may fail without indication, while others may return a null reference error where a more specific exception message might be warranted. This would be enough change to warrant a new PR however.

For this specific instance I want an exception thrown if a user fails to create -- not as vital if it does so when a user fails to drop, as my changes will reflect.

src/SmoFacade/User.cs
alexirion10: Can you please test and confirm this?  "understanding" from documentation is far from working IRL

mja9: I've tested it -- it successfully drops the user from the database.

src/SmoFacade/Database.cs
```diff
@@ -50,6 +50,29 @@ public void DropAssociatedLogins()
...
+      var user = Users.SingleOrDefault(u => u.Name.Equals(userProperties.Name, StringComparison.InvariantCultureIgnoreCase));
+      if (user == null) {
+        var smoUser = new Microsoft.SqlServer.Management.Smo.User(_database, userProperties.Name)
+          { Login = userProperties.LoginName };
+        smoUser.Create();
+        user = new User(smoUser, _server, this);
+
+        foreach (var role in userProperties.Roles) user.AddRole(role);
+        user.GrantPermission(userProperties.Permissions);
+      }
```
alexirion10, line 64: Can you code format this?  Inconsistent use of `{}` on new lines, same lines, or none

AdityaNPL (approved):
After the changes that Alex has requested, this looks good to go!

alexirion10:
can you update the branch?

alexirion10 (approved):

Issue #91, opened 2022-03-15
Server::Restore cannot restore striped backups


alexirion10:
striped backup's fail because the forloop attempts to restore on each backupMedia file

AdityaNPL:
Looks like something like this could work if we were using T-SQL
```sql
RESTORE DATABASE [db] FROM  
DISK = N'C:\db_1.bak',  
DISK = N'C:\db_2.bak',  
DISK = N'C:\db_3.bak',  
DISK = N'C:\db_4.bak',  
WITH  FILE = 1,  NOUNLOAD,  STATS = 10
GO
```
https://www.mssqltips.com/sqlservertip/1941/striping-sql-server-database-backups/

But looks like we are using the `Restore` class - need to dig into that
https://github.com/factset/AgDatabaseMove/blob/48bb272c233e3b820518ac28e898aa2f6631674a/src/SmoFacade/Server.cs#L225

alexirion10:
I wouldn't say "cluster" is the right description here, perhaps group?  but yes, that example TSQL looks correct.
I ran into a problem where if the full backup is stripped, but you only restore a part of the striped backup, it errors out

PR #101, opened 2022-04-26
feat(striped-backup): supporting striped backups

AdityaNPL:
Resolves: https://github.com/factset/AgDatabaseMove/issues/91

---

- Tested on a local MSSQL server and it could combine striped backups together in the chain and then use that chain to restore the DB
- Unit tests pass

### DIFF
--- a/src/BackupMetadata.cs
+++ b/src/BackupMetadata.cs
@@ -7,19 +7,9 @@ namespace AgDatabaseMove
 
   public class StripedBackupEqualityComparer : IEqualityComparer<BackupMetadata>
   {
-    private static StripedBackupEqualityComparer _instance = null;
+    private static readonly Lazy<StripedBackupEqualityComparer> s_instance = new Lazy<StripedBackupEqualityComparer>(() => new StripedBackupEqualityComparer());
     private StripedBackupEqualityComparer() { }
-
-    public static StripedBackupEqualityComparer Instance {
-      get
-      {
-        if (_instance == null)
-        {
-          _instance = new StripedBackupEqualityComparer();
-        }
-        return _instance;
-      }
-    }
+    public static StripedBackupEqualityComparer Instance => s_instance.Value;
 
     public bool Equals(BackupMetadata x, BackupMetadata y)
     {
@@ -51,22 +41,11 @@ namespace AgDatabaseMove
   /// </summary>
   public class BackupMetadataEqualityComparer : IEqualityComparer<BackupMetadata>
   {
-    private readonly StripedBackupEqualityComparer _stripedBackupEqualityComparer = StripedBackupEqualityComparer.Instance;
+    private static readonly StripedBackupEqualityComparer _stripedBackupEqualityComparer = StripedBackupEqualityComparer.Instance;
 
-    private static BackupMetadataEqualityComparer _instance = null;
+    private static readonly Lazy<BackupMetadataEqualityComparer> s_instance = new Lazy<BackupMetadataEqualityComparer>(() => new BackupMetadataEqualityComparer());
     private BackupMetadataEqualityComparer() { }
-
-    public static BackupMetadataEqualityComparer Instance
-    {
-      get
-      {
-        if (_instance == null)
-        {
-          _instance = new BackupMetadataEqualityComparer();
-        }
-        return _instance;
-      }
-    }
+    public static BackupMetadataEqualityComparer Instance => s_instance.Value;
 
     public bool Equals(BackupMetadata x, BackupMetadata y)
     {


src/BackupMetadata.cs
```diff
@@ -39,12 +41,17 @@ public int GetHashCode(BackupMetadata obj)
...
+    {
+      PhysicalDeviceNames = new List<string>();
+    }
     public decimal CheckpointLsn { get; set; }
     public decimal DatabaseBackupLsn { get; set; }
     public string DatabaseName { get; set; }
     public decimal FirstLsn { get; set; }
     public decimal LastLsn { get; set; }
-    public string PhysicalDeviceName { get; set; }
+    public List<string> PhysicalDeviceNames { get; set; }
```
alexirion10, line 54: I see where you're going with this... I prefer you not change this to a `List`.  This class is meant to represent 1 return row of data from `msdb.dbo.backupset`.
If the goal is to bundle the striped backups together, I suggest you create a new data type joins striped records into a list where needed

AdityaNPL: Yeah...I was debating on which approach to use as well

I feel `BackupMetadata` represents an instance of a backup, not a row of data from the query. That's why I changed the backupfile field from a single string to a list.
I can see previously, that it was synonymous with a row of data returned from the query too.

I'd like to argue, this is still a better approach than creating a new class to represent Striped backups.

However, if we do need to - most of the properties and behaviors will be the same.
So I'm thinking...it might be better to create it as a subclass of `BackupMetadata`?

src/StripedBackupSet.cs
```diff
@@ -0,0 +1,37 @@
...
+    public StripedBackupSet(IEnumerable<BackupMetadata> backups, BackupMetadata backupToMatch)
+    {
+      var stripes = backups.Where(backup => AreTwoBackupsStriped(backup, backupToMatch));
+      StripedBackups = new List<BackupMetadata>(stripes);
+    }
+
+    public static bool AreTwoBackupsStriped(BackupMetadata backup, BackupMetadata backupToMatch)
+    {
+      return new BackupMetadataEqualityComparer().EqualsExceptForPhysicalDeviceName(backup, backupToMatch);
+    }
```
alexirion10, line 22: This does not need to be a standalone static func

src/StripedBackupSet.cs
```diff
@@ -0,0 +1,37 @@
...
+
+  public class StripedBackupSet
+  {
+    public IEnumerable<BackupMetadata> StripedBackups { get; private set; }
+
+    public StripedBackupSet(IEnumerable<BackupMetadata> backups, BackupMetadata backupToMatch)
+    {
+      var stripes = backups.Where(backup => AreTwoBackupsStriped(backup, backupToMatch));
+      StripedBackups = new List<BackupMetadata>(stripes);
+    }
```
alexirion10, line 17: This should be a private constructor; ie. nothing should be calling this outside of this class itself

src/StripedBackupSet.cs
```diff
@@ -0,0 +1,37 @@
...
+    public static IEnumerable<StripedBackupSet> GetStripedBackupSetChain(IEnumerable<BackupMetadata> backups)
+    {
+      List<StripedBackupSet> chain = new List<StripedBackupSet>(); 
+      while(backups.Count() > 0)
+      {
+        var backup = backups.First();
+        var stripedBackupSet = new StripedBackupSet(backups, backup);
+        chain.Add(stripedBackupSet);
+        backups = backups.Except(stripedBackupSet.StripedBackups, new BackupMetadataEqualityComparer());
+      }
```
alexirion10, line 33: Outer loop here makes sense... but I don't care for removing items from the list and passing the list around constantly (copying data).
Why don't you order the backup list, I think it might already be ordered....
With ordering any striped backup is always going to be the next element... roll them up until you find a non-matching `BackupMetadata`.  I think this can be accomplished with a LINQ GroupBy clause

src/AgDatabaseMove.cs
```diff
@@ -73,8 +73,10 @@ public decimal Move(decimal? lastLsn = null)
 
       if(!backupList.Any())
         throw new BackupChainException("No backups found to restore");
+      
+      var stripedBackupSetChain = StripedBackupSet.GetStripedBackupSetChain(backupList.ToList());
```
alexirion10, line 77: `backupList` is already an `IList<BackupMetadata>` type.  Why call `.ToList()` on it?

src/StripedBackupSet.cs
```diff
@@ -0,0 +1,37 @@
...
+using System.Text;
+using System.Linq;
+
+namespace AgDatabaseMove
+{
+
+  public class StripedBackupSet
+  {
+    public IEnumerable<BackupMetadata> StripedBackups { get; private set; }
+
```
alexirion10, line 12: I think this makes alot more sense to be in the `BackupChain` class itself

src/SmoFacade/Server.cs
```diff
@@ -192,38 +192,63 @@ public Database Database(string dbName)
...
+        AddBackupDeviceItemsToRestore(restore, stripedBackupSet);
 
         var defaultFileLocations = DefaultFileLocations();
-        if(defaultFileLocations != null) {
-          restore.RelocateFiles.Clear();
-          var fileList = policy.Execute(() => restore.ReadFileList(_server).AsEnumerable());
-          foreach(var file in fileList) {
-            var physicalName = (string)file["PhysicalName"];
-            var fileName = GetFileName(physicalName);
+        if (defaultFileLocations != null)
```
alexirion10, line 200: Looks like this query actually only needs to be done 1 time, not on each loop iteration.
Can you move it out of the loop?
https://github.com/factset/AgDatabaseMove/blob/a89418883f5c3275b561b6edd943f6f0bf76dbed/src/SmoFacade/Server.cs#L123

src/SmoFacade/Server.cs
```diff
@@ -178,11 +178,11 @@ public Database Database(string dbName)
     ///   wait
     /// </param>
     /// <param name="fileRelocation">Option for renaming files during the restore.</param>
-    public void Restore(IEnumerable<BackupMetadata> backupOrder, string databaseName,
+    public void Restore(IEnumerable<StripedBackupSet> stripedBackupSetChain, string databaseName,
       Func<int, TimeSpan> retryDurationProvider,
       Func<string, string> fileRelocation = null)
     {
-      var policy = Policy
+      var retryPolicy = Policy
```
alexirion10, line 185: this can probably be better served as a class static variable so it doesn't have to be passed around as a parameter

tests/AgDatabaseMove.Unit/StripedBackupSetTests.cs
```diff
@@ -0,0 +1,50 @@
...
+  public class StripedBackupSetTests
+  {
+
+    [Fact]
+    public void CombinesStripedBackups()
+    {
+      var backupChain = BackupOrder.GetBackupListWithStripes();
+      var stripedBackupSetChain = StripedBackupSet.GetStripedBackupSetChain(backupChain);
+
+      Assert.Equal(backupChain.Count/2, stripedBackupSetChain.Count());
```
alexirion10, line 18: Instead of `Assert.Equal(backupChain.Count/2`, ..)` I think a better assert would be
`Assert.Equal(backupChain.DistinctBy(...).Count, ...)`

DistinctBy or GroupBy:
https://stackoverflow.com/questions/6231360/linq-query-to-return-distinct-field-values-from-list-of-objects

tests/AgDatabaseMove.Unit/StripedBackupSetTests.cs
```diff
@@ -0,0 +1,50 @@
...
+      foreach (var stripedBackupSet in stripedBackupSetChain)
+      {
+        var stripes = stripedBackupSet.StripedBackups.ToList();
+        var backup = stripes.First();
+        for (int i = 1; i < stripes.Count; i++)
+        {
+          var otherBackup = stripes[i];
+          
+          Assert.True(new BackupMetadataEqualityComparer().EqualsExceptForPhysicalDeviceName(backup, otherBackup));
+          Assert.NotEqual(backup, otherBackup, new BackupMetadataEqualityComparer());
```
alexirion10, line 45: Would be nice to not call `new BackupMetadataEqualityComparer()` so many times.  Can you resue 1 instance?

alexirion10 (changes requested):
On the right track, some improvement ideas below

src/StripedBackupSet.cs
AdityaNPL: Ya actually `GroupBy` might be best. Then we just have to change those groups into `StripedBackupSet`
Will take a look

src/AgDatabaseMove.cs
AdityaNPL: I'll take a look, might be a remnant from when had the parameter as `List` type instead of `IEnumerable`

src/SmoFacade/Server.cs
AdityaNPL: The `restore()` function takes in a `Func<int, TimeSpan> retryDurationProvider` parameter and we use that to create this policy.

I'm wondering if we can pass it in the constructor of the `Server` class instead and then make `retryPolicy` a member variable?
Or maybe just have a static retryDuration?

tests/AgDatabaseMove.Unit/StripedBackupSetTests.cs
AdityaNPL: Will take a look

src/StripedBackupSet.cs
```diff
@@ -0,0 +1,34 @@
...
+        {
+          b.FirstLsn,
+          b.LastLsn,
+          b.CheckpointLsn,
+          b.DatabaseBackupLsn,
+          b.BackupType,
+          b.DatabaseName
+        })
+        .Select(group => new StripedBackupSet(group.ToList()));
+      return chain;
```
AdityaNPL, line 31: We can make this even better if we create a new custom `IEqualityComparer` class - maybe named `StripedBackupMetadataComparator`
@alexirion10 what do you think?

```csharp
var chain = backups
        .GroupBy(b => b, new StripedBackupMetadataComparator())
        .Select(group => new StripedBackupSet(group.ToList()));
```

tests/AgDatabaseMove.Unit/StripedBackupSetTests.cs
AdityaNPL: Since we have changed the code to use `GroupBy`, having the unit test use `GroupBy` too doesn't make sense anymore.
We won't be checking anything useful.

This unit test will check whether our grouping logic is correct. So I feel we should leave it as is.

I think I some extra checks to make sure each striped backup set is distinct but each backup within a specific striped backup set is the same?

Let me know what you think?

src/StripedBackupSet.cs
alexirion10: yes

src/SmoFacade/Server.cs
alexirion10: ah.  forgot that it depends on the parameter. Leave it as is

tests/AgDatabaseMove.Unit/StripedBackupSetTests.cs
alexirion10: `Assert.Equal(backupChain.Count/2, ...)` is a test that is going to fail as soon as records are added or removed from the backup list.
I'd like you to write a test that would continue to pass if another record were added to the backup meta list

src/StripedBackupSet.cs
alexirion10: Either way sounds fine.  
What is the point of https://github.com/factset/AgDatabaseMove/blob/master/src/BackupMetadata.cs#L13 then?  Can you consolidate into a single comparator?

I don't think you need to call `ToList` here.  You're passing in an IEnumerable

AdityaNPL: > Can you consolidate into a single comparator?

Yes that function will now be the `Equals()` function of the new comparator

> I don't think you need to call ToList here. You're passing in an IEnumerable

the `group` is an `IGrouping` which is an `IEnumerable` - correct
But it stores it as a dict if I don't do `ToList()`. But, I guess that should still be ok as long as it can enumerate over the stripes. 
(One of the benefits of using `IEnumerable`)
ref: https://docs.microsoft.com/en-us/dotnet/api/system.linq.igrouping-2?view=net-5.0

AdityaNPL: Created a new `StripedBackupEqualityComparer`

tests/AgDatabaseMove.Unit/StripedBackupSetTests.cs
AdityaNPL: `Distinct()` works to get the correct count
(Tested by increasing the stripes in each stripedSet)

src/BackupMetadata.cs
```diff
@@ -20,21 +17,10 @@ public bool EqualsExceptForPhysicalDeviceName(BackupMetadata x, BackupMetadata y
              x.DatabaseBackupLsn == x.DatabaseBackupLsn;
```
mja9, line 17: Not part of the changes here but something I noticed -- Should this be `x.DatabaseBackupLsn == y.DatabaseBackupLsn` or is there a reason I'm missing for why it's written this way?

tests/AgDatabaseMove.Unit/StripedBackupSetTests.cs
```diff
@@ -0,0 +1,56 @@
...
+    {
+      var backupChain = BackupOrder.GetBackupListWithStripes();
+
+      var stripedBackupSetChain = StripedBackupSet.GetStripedBackupSetChain(backupChain);
+
+      Assert.Equal(backupChain.Distinct(stripedBackupComparer).Count(), stripedBackupSetChain.Count());
+    }
+
+    [Fact]
+    public void DoesntCombineNonStripedBackups()
```
mja9, line 26: Tests look good -- I see from your comments you tested on a local MSSQL server to verify it could restore from striped backups. 

Did you happen to verify that restoring without a striped backup chain works as intended as well?

mja9 (commented):
Looking good to me! Just a couple questions below.

src/BackupMetadata.cs
AdityaNPL: ah, great catch!
Was a typo, I'll fix that

tests/AgDatabaseMove.Unit/StripedBackupSetTests.cs
AdityaNPL: I think I did run that once, but not recently after the new changes
Will give it a go

AdityaNPL: Just normal backups still work
and so does mixing and matching 👍🏽

src/BackupMetadata.cs
```diff
@@ -43,6 +29,28 @@ public int GetHashCode(BackupMetadata obj)
...
     }
+
+  }
+
+  /// <summary>
+  /// Two BackupMetadatas are the same, if they are like striped backups but also have the same `PhysicalDeviceName`
+  /// </summary>
+  public class BackupMetadataEqualityComparer : IEqualityComparer<BackupMetadata>
+  {
+    private readonly StripedBackupEqualityComparer _stripedBackupEqualityComparer = new StripedBackupEqualityComparer();
```
alexirion10, line 40: I might suggest that you look into creating a static, singleton, instance of this class.
You call `new StripedBackupEqualityComparer` several times that seem unnecessary.  I think this would all work with 1 instance of the class. Let me know

AdityaNPL: Agreed! Made them a singleton

src/BackupMetadata.cs
```diff
@@ -2,39 +2,39 @@ namespace AgDatabaseMove
...
+    private StripedBackupEqualityComparer() { }
+
+    public static StripedBackupEqualityComparer Instance {
+      get
+      {
+        if (_instance == null)
+        {
+          _instance = new StripedBackupEqualityComparer();
+        }
+        return _instance;
```
slin5299, line 20: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-8.0/null-coalescing-assignment

I think this can be simplified using `??=`. Something like
```
_instance ??= new StripedBackupEqualityComparer();
return _instance ;
```

or even 
` return _instance ??= new StripedBackupEqualityComparer();`

slin5299 (approved):
Looks good to me, just one suggestion

mja9 (approved):

src/BackupMetadata.cs
alexirion10: yeah... this doesn't even need to be a public property with a `Getter`

```
private static readonly Lazy<StripedBackupEqualityComparer> s_instance = new Lazy<StripedBackupEqualityComparer>();

    private StripedBackupEqualityComparer() { }

    public static StripedBackupEqualityComparer Instance => s_instance.Value;
```

AdityaNPL: ah cool, didn't know you could make anything a lazy initialisation - will try it out
https://docs.microsoft.com/en-us/dotnet/api/system.lazy-1?view=net-6.0

PR #102, opened 2022-04-28
feat(restore): logical file rename

mja9:
Made changes to `Restore` to rename logical files using `fileRelocator` when passed in.

Tested in separate console application, verified logical names had changed in SSMS

### DIFF
--- a/src/AgDatabase.cs
+++ b/src/AgDatabase.cs
@@ -34,6 +34,7 @@ namespace AgDatabaseMove
     void Restore(IEnumerable<BackupMetadata> backupOrder, Func<int, TimeSpan> retryDurationProvider,
       Func<string, string> fileRelocation = null);
 
+    void RenameLogicalFileName(Func<string, string> fileRenamer);
     void AddLogin(LoginProperties login);
     IEnumerable<LoginProperties> AssociatedLogins();
     void DropLogin(LoginProperties login);
--- a/src/AgDatabase.cs
+++ b/src/AgDatabase.cs
@@ -120,7 +120,7 @@ namespace AgDatabaseMove
 
     public void RenameLogicalFileName(Func<string, string> fileRenamer)
     {
-      _listener.Primary.LogicalFileRename(Name, fileRenamer);
+      _listener.Primary.RenameLogicalFileName(Name, fileRenamer);
     }
 
     /// <summary>
--- a/src/AgDatabase.cs
+++ b/src/AgDatabase.cs
@@ -118,6 +118,11 @@ namespace AgDatabaseMove
       _listener.ForEachAgInstance(s => s.Restore(backupOrder, Name, retryDurationProvider, fileRelocation));
     }
 
+    public void RenameLogicalFileName(Func<string, string> fileRenamer)
+    {
+      _listener.Primary.LogicalFileRename(Name, fileRenamer);
+    }
+
     /// <summary>
     ///   Builds a list of recent backups from msdb on each AG instance.
     /// </summary>
--- a/src/SmoFacade/Server.cs
+++ b/src/SmoFacade/Server.cs
@@ -18,15 +18,17 @@ namespace AgDatabaseMove.SmoFacade
   public class Server : IDisposable
   {
     private readonly string _connectionString;
+    private readonly string _agName;
     internal readonly string _credentialName;
     internal readonly Microsoft.SqlServer.Management.Smo.Server _server;
 
-    public Server(string connectionString, string credentialName = null)
+    public Server(string connectionString, string credentialName = null, string agName = "")
     {
       _connectionString = connectionString;
       SqlConnection = new SqlConnection(_connectionString);
       _server = new Microsoft.SqlServer.Management.Smo.Server(new ServerConnection(SqlConnection));
       _credentialName = credentialName;
+      _agName = agName;
     }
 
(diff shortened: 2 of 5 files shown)


src/SmoFacade/Server.cs
```diff
@@ -224,7 +224,29 @@ public Database Database(string dbName)
...
+
+    private void LogicalFileRename(string databaseName, Func<string, string> fileRelocation)
+    {
+      var db = Database(databaseName);
+      if (db.Restoring)
+        db.RestoreWithRecovery();
+
+      foreach (FileGroup fileGroup in (SmoCollectionBase)db._database.FileGroups)
+      {
+        DataFile[] array = new DataFile[fileGroup.Files.Count];
```
AdityaNPL, line 241: Please rename these arrays to something more meaningful.
`dataFiles` and `logFiles` seems to work?

src/SmoFacade/Server.cs
```diff
@@ -224,7 +224,29 @@ public Database Database(string dbName)
 
         policy.Execute(() => restore.SqlRestore(_server));
         restore.Devices.Remove(backupDeviceItem);
+
+        if(fileRelocation != null)
+          LogicalFileRename(databaseName, fileRelocation);
+      }
```
AdityaNPL, line 230: I'm guessing we are doing this after each restore because it is a clean up action?
(and something we don't want to happen if the restore fails?)

Which begs the question what happens if one of iteration fails (or maybe it fails on one server).
Can we still re-run the `move()`/`restore()` function again?

alexirion10: This should be done after the entire restore completes, not on each loop iteration.
if the restore fails & is retried, the db is overwritten

src/SmoFacade/Server.cs
```diff
@@ -224,7 +224,29 @@ public Database Database(string dbName)
...
+        if(fileRelocation != null)
+          LogicalFileRename(databaseName, fileRelocation);
+      }
+    }
+
+    private void LogicalFileRename(string databaseName, Func<string, string> fileRelocation)
+    {
+      var db = Database(databaseName);
+      if (db.Restoring)
+        db.RestoreWithRecovery();
```
alexirion10, line 237: Don't want to have this logic in here... if you restore the database here, the rest of the chain won't be able to restore haha.

This function should only be called after the full chain in restored & finalized

src/SmoFacade/Server.cs
mja9: Whoops, yes this was intended to be placed _after_ the loop. I have a local branch where I'm testing all the changes I've made together. I must've mistyped when copying this addition over from there!

src/SmoFacade/Server.cs
```diff
@@ -225,6 +225,28 @@ public Database Database(string dbName)
...
+
+      if (fileRelocation != null)
+        LogicalFileRename(databaseName, fileRelocation);
+    }
+
+    private void LogicalFileRename(string databaseName, Func<string, string> fileRelocation)
+    {
+      var db = Database(databaseName);
+      if (db.Restoring)
+        db.RestoreWithRecovery();
```
alexirion10, line 237: We don't want to finalize the `Restore` operation here.  

Notice how we finalize based on this property, which could be... false!
https://github.com/factset/AgDatabaseMove/blob/master/src/AgDatabaseMove.cs#L83

Perhaps this is here because you can only rename the logicalfilename if the DB is restored?  
To note, this rename logic doesn't necessarily need to be deeply nested here in the restore code.  
Now that this is a standalone function on `Server` - make it public/internal... and call it from other classes where applicable (ie. [here](https://github.com/factset/AgDatabaseMove/blob/master/src/AgDatabaseMove.cs#L85) )

alexirion10: I can't drop our FDS code in this public repo... take a look at example usage in  FactSet.Sqlis.Core/Move.cs#L44-L68

alexirion10 (changes requested):

src/SmoFacade/Server.cs
mja9: Yes, I threw it in there because you can only access the file names once the DB is restored. 

I'll move forward with making the function public. Given that information, I think I'll throw an exception rather than finalize the restore on L237 to ensure finalizing is done intentionally.

src/AgDatabase.cs
```diff
@@ -118,6 +118,11 @@ public void LogBackup()
       _listener.ForEachAgInstance(s => s.Restore(backupOrder, Name, retryDurationProvider, fileRelocation));
     }
 
+    public void RenameLogicalFileName(Func<string, string> fileRenamer)
+    {
+      _listener.Primary.RenameLogicalFileName(Name, fileRenamer);
+    }
+
```
AdityaNPL, line 125: I see what we are doing, that way the caller can call this separately if they need to.

But we should be able to call this from the `Move()` function too right?
https://github.com/factset/AgDatabaseMove/blob/b3f444b7d0e85caf98a0deb9e0528067e4b4068e/src/AgDatabaseMove.cs#L54

AdityaNPL (approved):
LGTM!
Just a quick Q  below

slin5299 (approved):
looks good me

alexirion10 (approved):

PR #103, opened 2022-05-03
fix(backup-chain): next log backup can have the same `LastLsn`

AdityaNPL:
(It looks like it made loads of white space changes automatically - pls ignore this while reviewing.)

I came across an instance where the `LastLsn` was the same for 2 backups. Hence, this check was filtering the log backup out.

image

### DIFF
--- a/src/BackupChain.cs
+++ b/src/BackupChain.cs
@@ -92,7 +92,9 @@ namespace AgDatabaseMove
     {
       // also gets all the stripes of the next backup
       return backups.Where(b => b.BackupType == BackupFileTools.BackupType.Log &&
-                                prevBackup.LastLsn >= b.FirstLsn && prevBackup.LastLsn <= b.LastLsn);
+                                prevBackup.LastLsn >= b.FirstLsn && 
+                                prevBackup.LastLsn <= b.LastLsn &&
+                                !new BackupEqualityComparer().Equals(prevBackup, b));
     }
 
     private static bool IsValidFilePath(BackupMetadata meta)
--- a/src/BackupChain.cs
+++ b/src/BackupChain.cs
@@ -94,7 +94,7 @@ namespace AgDatabaseMove
       return backups.Where(b => b.BackupType == BackupFileTools.BackupType.Log &&
                                 prevBackup.LastLsn >= b.FirstLsn && 
                                 prevBackup.LastLsn <= b.LastLsn &&
-                                !new BackupEqualityComparer().Equals(prevBackup, b));
+                                !new BackupMetadataEqualityComparer().EqualsExceptForPhysicalDeviceName(prevBackup, b));
     }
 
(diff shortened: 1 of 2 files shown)


src/BackupChain.cs
```diff
@@ -54,51 +54,51 @@ private BackupChain(IList<BackupMetadata> recentBackups)
...
+      // get all the stripes of this backup
+      return diffBackupsOrdered.Where(diffBackup => diffBackup.LastLsn == targetLastLsn);
+    }
+
+    private static IEnumerable<BackupMetadata> NextLogBackup(IEnumerable<BackupMetadata> backups,
+      BackupMetadata prevBackup)
+    {
+      // also gets all the stripes of the next backup
+      return backups.Where(b => b.BackupType == BackupFileTools.BackupType.Log &&
+                                prevBackup.LastLsn >= b.FirstLsn && prevBackup.LastLsn <= b.LastLsn);
```
AdityaNPL, line 95: I'm wondering, if this is a simpler check?
```csharp
prevBackup.FirstLsn <= b.FirstLsn && prevBackup.LastLsn <= b.LastLsn
```

alexirion10: no, the point of chaining is to check first_lsn vs last_lsn

alexirion10 (commented):
Not sure what scenario caused this, but ok.  Take a look at the failing unit test plz

AdityaNPL:
Unit test:
> System.OutOfMemoryException : Exception of type 'System.OutOfMemoryException' was thrown.

huh, that's weird - checking

AdityaNPL:
uggh, infinite loop while building the chain cause it's matching the same log backup over and over again

src/BackupMetadata.cs
```diff
@@ -5,31 +5,45 @@ namespace AgDatabaseMove
   using SmoFacade;
 
 
-  /// <summary>
-  ///   Occasionally we wind up with the same entry for a backup on multiple instance's msdb.
-  ///   For now we'll consider these backups to be equal despite their file location,
-  ///   but perhaps there's value in being able to look for the file in multiple locations.
-  /// </summary>
-  public class BackupMetadataEqualityComparer : IEqualityComparer<BackupMetadata>
+  public class BackupEqualityComparer : IEqualityComparer<BackupMetadata>
```
AdityaNPL, line 8: Needed to borrow this from the other PR to fix the issue here:
https://github.com/factset/AgDatabaseMove/pull/101/files#diff-0b43e3447245b624777b212abcd1a5a74ed28b239eb4797e66e7ce06b3288e62R8

AdityaNPL:
Fixed the infinite loop by making sure it isn't getting the same backup as the last one

---

Tested on a MSSQL server with the lastLSNs equal for two of the backups
Also the unit tests should run fine now

alexirion10 (approved):

PR #104, opened 2022-05-05
feat(file): size-growth

mja9:
Changes made to alter data file max size and growth rate, and log file growth rate.

Tested on sandbox database and verified growth rates and max size changed accordingly in SSMS.

### DIFF
--- a/src/SmoFacade/Database.cs
+++ b/src/SmoFacade/Database.cs
@@ -45,10 +45,44 @@ namespace AgDatabaseMove.SmoFacade
     {
       var logins = Users.Where(u => u.Login != null && u.Login.Name != "sa")
         .Select(u => u.Login.Properties());
-      foreach(var login in logins)
-        _server.DropLogin(new LoginProperties {
+      foreach (var login in logins)
+        _server.DropLogin(new LoginProperties
+        {
           Name = login.Name
         });
+    }
+
+    public User AddUser(UserProperties userProperties)
+    {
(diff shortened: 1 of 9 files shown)


src/SmoFacade/Database.cs
```diff
@@ -14,6 +14,8 @@ public class Database
   {
     internal readonly Microsoft.SqlServer.Management.Smo.Database _database;
     private readonly Server _server;
+    private const int GB_TO_KB = 1048576;
```
mja9, line 17: @alexirion10 Since the use for these is confined to `Database.cs` and since they're well-known values, I think a `private const` is more appropriate than `public static readonly`.

AdityaNPL: Looks like there is a nuget package we could use instead, but I think it's owned by just one person (and not an organization) - so more risky

https://www.nuget.org/packages/ByteSize/
https://github.com/omar/ByteSize

alexirion10: I don't see the need for an external nuget package when this can simply be a `const int` in here

src/SmoFacade/Database.cs
```diff
@@ -195,6 +197,35 @@ public void MultiUserMode()
       _database.Alter(TerminationClause.RollbackTransactionsImmediately);
     }
 
+    public void SetSizeLimit(int maxGB)
+    {
```
alexirion10, line 201: Can you make this one `MB` for consistency with the other two?

alexirion10 (approved):
good to go - take a look at the comment above for consistency

PR #105, opened 2022-05-07
fix(backups): no full backup

AdityaNPL:
- [x] test on a test DB

### DIFF
--- a/src/AgDatabase.cs
+++ b/src/AgDatabase.cs
@@ -142,7 +142,7 @@ namespace AgDatabaseMove
 
       // find all backups in that chain
       if (fullBackupLsnBag.Count == 0)
-        throw new Exception("Could not find any full backups"); 
+        throw new Exception($"Could not find any full backups for DB '{Name}'"); 
 
       var databaseBackupLsn = fullBackupLsnBag.Max();
       var bag = new ConcurrentBag<BackupMetadata>();
--- a/src/AgDatabase.cs
+++ b/src/AgDatabase.cs
@@ -131,9 +131,19 @@ namespace AgDatabaseMove
     {
       // find most recent full backup LSN across all replica servers
       var fullBackupLsnBag = new ConcurrentBag<decimal>();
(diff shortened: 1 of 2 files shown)


src/AgDatabase.cs
```diff
@@ -131,9 +131,19 @@ public List<BackupMetadata> RecentBackups()
...
+        try
+        {
+          fullBackupLsnBag.Add(s.Database(Name).MostRecentFullBackupLsn());
+        }
+        catch { }
+      });
 
       // find all backups in that chain
+      if (fullBackupLsnBag.Count == 0)
+        throw new Exception("Could not find any full backups");
```
lotsofspots, line 145: Do we want to append the Name here?

AdityaNPL: can do

src/AgDatabase.cs
```diff
@@ -131,9 +131,19 @@ public List<BackupMetadata> RecentBackups()
...
+      {
+        try
+        {
+          fullBackupLsnBag.Add(s.Database(Name).MostRecentFullBackupLsn());
+        }
+        catch { }
+      });
 
       // find all backups in that chain
+      if (fullBackupLsnBag.Count == 0)
```
mja9, line 144: Is there a `fullBackupLsnBag.IsEmpty()`?

Kaneraz (approved):
Looks correct. 

When throwing a specific exception I prefer to make a new type, and catch it rather than just throw exception. 

Also, it can be expected that a backup may not exist on each node, so the interface should change to be nullable. Save the exceptions for when something unexpected happens.

src/AgDatabase.cs
AdityaNPL: yup, looks like there is - changed it 👍🏽

AdityaNPL:
@Kaneraz - I agree, maybe we can return `decimal?` or some other nullable value instead.
I need this fix right now so will merge it in and follow up with a refactor tomorrow.

PR #106, opened 2022-05-09
refactor(backups): MostRecentFullBackupLsn can return null

AdityaNPL:
Continuation of https://github.com/factset/AgDatabaseMove/pull/105#issuecomment-1120393596

Tested:

- [X] Unit tests
- [ ] Test DB


alexirion10:
ahh... yes!  Thanks for picking this up, merging this in, and bumping the package

PR #108, opened 2022-06-02
perf(AgDatabase): Delete

mja9:
I'd been experiencing issues with `Delete` while testing the changes I was making for creating databases. Specifically, despite the retry logic in place the method seemed to be throwing this [exception](https://github.com/factset/AgDatabaseMove/blob/43a0983f687358a669cf74ddb47b383ff21276a3/src/SmoFacade/Database.cs#L287-L289) on a high volume of invocations.

As I was investigating the cause I found this [post](https://stackoverflow.com/questions/27950110/smo-equivalent-to-set-hadr-off), which led me to believe our logic for leaving an availability group on secondary servers may be taking a needlessly lengthy amount of time.

Tested with the local `AgDatabaseMove.Cli` and found the database was able to drop successfully without throwing, all within ~4000 ms versus the old execution time ~30000 ms (when the method worked on the first invocation). 

I believe we currently have an additional retry policy surrounding the call to `Delete()` in one of our repositories. This change should remove the need to nest retry policies altogether.

### DIFF
--- a/src/AgDatabase.cs
+++ b/src/AgDatabase.cs
@@ -99,8 +99,15 @@ namespace AgDatabaseMove
     /// </summary>
     public void Delete()
     {
+      _listener.ForEachAgInstance((s, ag) => {
+        if (!ag.IsPrimaryInstance)
+        {
+          ag.Remove(Name);
+          s.Database(Name)?.Drop();
+        }
+      });
       _listener.AvailabilityGroup.Remove(Name);
-      _listener.ForEachAgInstance(s => s.Database(Name)?.Drop());
+      _listener.Primary.Database(Name)?.Drop();
     }
 
     /// <summary>
--- a/src/SmoFacade/AvailabilityGroup.cs
+++ b/src/SmoFacade/AvailabilityGroup.cs
@@ -57,7 +57,13 @@ namespace AgDatabaseMove.SmoFacade
 
     public void Remove(string dbName)
     {
-      _availabilityGroup.AvailabilityDatabases[dbName]?.Drop();
+      if (IsPrimaryInstance)
+      {
+        _availabilityGroup.AvailabilityDatabases[dbName]?.Drop();
+      } else
+      {
+        _availabilityGroup.AvailabilityDatabases[dbName]?.LeaveAvailabilityGroup();
+      }
     }
 
     public bool IsInitializing(string dbName)


src/AgDatabase.cs
```diff
@@ -99,8 +99,15 @@ public bool Exists()
...
+      _listener.ForEachAgInstance((s, ag) => {
+        if (!ag.IsPrimaryInstance)
+        {
+          ag.Remove(Name);
+          s.Database(Name)?.Drop();
+        }
+      });
       _listener.AvailabilityGroup.Remove(Name);
-      _listener.ForEachAgInstance(s => s.Database(Name)?.Drop());
+      _listener.Primary.Database(Name)?.Drop();
```
AdityaNPL, line 110: I think all this can be simplified down to:

```csharp
_listener.ForEachAgInstance((s, ag) => {
    ag.Remove(Name);
    s.Database(Name)?.Drop();
});
```

This looks like it works even if executed on the primary first, tested it locally via this SQL command:
```sql
USE [master]

GO

ALTER AVAILABILITY GROUP [***(redacted)]
REMOVE DATABASE [aniraula_2];

GO

DROP DATABASE [aniraula_2];

Go

```

src/SmoFacade/AvailabilityGroup.cs
```diff
@@ -57,7 +57,13 @@ public void JoinPrimary(string dbName)
...
     public void Remove(string dbName)
     {
-      _availabilityGroup.AvailabilityDatabases[dbName]?.Drop();
+      if (IsPrimaryInstance)
+      {
+        _availabilityGroup.AvailabilityDatabases[dbName]?.Drop();
+      } else
+      {
+        _availabilityGroup.AvailabilityDatabases[dbName]?.LeaveAvailabilityGroup();
+      }
```
AdityaNPL, line 66: I haven't looked into what the diff b/w the two calls are yet and what `HADR` means, but just wondering before I do...
It looks like, in the old code, we weren't removing the DB from the secondary AG before trying to drop the DB on the secondary servers
Could it just be that which was causing the errors?

src/AgDatabase.cs
```diff
@@ -99,8 +99,15 @@ public bool Exists()
     /// </summary>
     public void Delete()
     {
+      _listener.ForEachAgInstance((s, ag) => {
+        if (!ag.IsPrimaryInstance)
+        {
+          ag.Remove(Name);
+          s.Database(Name)?.Drop();
```
alexirion10, line 106: as the stackoverflow post mentions, is it worth calling `.SuspendDataMovement();` as well?

alexirion10 (approved):
lgtm!  some comments above to mull over

src/AgDatabase.cs
mja9: This was my original implementation, but I was seeing the same slow performance while testing. It's been a few days, so I'll doublecheck and report back here what I find.

src/AgDatabase.cs
mja9: My gut feeling is yes, but I need to look into further. 
I'll add a more fleshed out rationale for why it is/isn't worthwhile once I've researched a bit more.

src/AgDatabase.cs
mja9: I haven't been able to figure out exactly why, but simplifying the call to the suggestion you made above results in the same exception being thrown. After doing some debugging, I think it may be because the availability database gets dropped on the primary before the call to `LeaveAvailabilityGroup()` on the secondary's.

So to better answer your other question below, I think the problem isn't just making sure we drop the db on the secondary ag before deleting the db, but making sure we drop the db on the secondary ag _before_ we do so on the primary.

src/AgDatabase.cs
mja9: From what I can tell after some researching, `SuspendDataMovement()` may not be necessary. The process of removing the database from the availability group _should_ stop the synchronization on the secondary's (and I have yet to see any reported errors for attempting to remove an availability database while data is synchronizing). The biggest difference I can see is that suspending HADR will keep the log queues intact, i.e. the primaries send queue will continue to build up logs until the database is removed and the corresponding queue is presumably destroyed.

alexirion10: > The process of removing the database from the availability group should stop the synchronization on the secondary's  
yes

> suspending HADR will keep the log queues intact, i.e. the primaries send queue will continue to build up logs until the database is removed and the corresponding queue is presumably destroyed.  
yes, that would make a lot of sense!  Since we are dropping the DB in this case, I think we are good to remove the DB from the AG without worrying about keeping the log queue :)

good to go!

PR #109, opened 2022-10-11
fix(backupchain): diff and log share last lsn

mja9:
I was running into an error with the backup chain construction on a sample of backups identical (in their order) to the example I've provided below as a unit test. I've included a fix in here that passes the provided unit test, but wanted to get additional eyes on this to verify the change shouldn't effect how backups are being ordered in other use cases.

Specifically, the variables used within the `while` loop would change as follows:

First Iteration
----
`orderedBackups `= [full, diff]
`prevBackup `= diff

`nextLog `= [log2, log1]

Second Iteration
----
`orderedBackups `= [full, diff, log2, log1]
`prevBackup `= log1

`nextLog `= [log2]

....

Such that the resulting `orderedBackups` = [full, diff, log2, log1, log2, ...]. This would result in a bug in which a striped backup chain built from these backups would incorrectly identify both log2 elements as part of a single striped backup instead of duplicates.

### DIFF
--- a/src/BackupChain.cs
+++ b/src/BackupChain.cs
@@ -30,7 +30,7 @@ namespace AgDatabaseMove
 
       var orderedBackups = MostRecentFullBackup(backups).ToList();
       orderedBackups.AddRange(MostRecentDiffBackup(backups, orderedBackups.First()));
-      orderedBackups.AddRange(NextLogBackupAfterDiff(backups, orderedBackups.Last()));
+      orderedBackups.AddRange(FirstBackupInChain(backups, orderedBackups.Last()));
 
       var prevBackup = orderedBackups.Last();
       IEnumerable<BackupMetadata> nextLogBackups;
@@ -91,23 +91,29 @@ namespace AgDatabaseMove
     private static IEnumerable<BackupMetadata> NextLogBackup(IEnumerable<BackupMetadata> backups,
       BackupMetadata prevBackup)
     {
-      // also gets all the stripes of the next backup
-      return backups.Where(b => b.BackupType == BackupFileTools.BackupType.Log &&
-                                prevBackup.LastLsn == b.FirstLsn &&
-                                !StripedBackupEqualityComparer.Instance.Equals(prevBackup, b))
-                    .OrderBy(b => b.LastLsn);
+     // also gets all the stripes of the next backup
+     return backups.Where(b => b.BackupType == BackupFileTools.BackupType.Log &&
+                               prevBackup.LastLsn == b.FirstLsn &&
+                               !StripedBackupEqualityComparer.Instance.Equals(prevBackup, b));
     }
 
-    private static IEnumerable<BackupMetadata> NextLogBackupAfterDiff(IEnumerable<BackupMetadata> backups, 
+    private static IEnumerable<BackupMetadata> FirstBackupInChain(IEnumerable<BackupMetadata> backups, 
      BackupMetadata lastDiff)
     {
-      return backups.Where(b => b.BackupType == BackupFileTools.BackupType.Log &&
+      var possibleLogs = backups.Where(b => b.BackupType == BackupFileTools.BackupType.Log &&
                                        lastDiff.LastLsn >= b.FirstLsn &&
                                        lastDiff.LastLsn <= b.LastLsn &&
-                                       !StripedBackupEqualityComparer.Instance.Equals(lastDiff, b))
-                    .GroupBy(b => b, StripedBackupEqualityComparer.Instance)
-                    .OrderBy(b => b.First().LastLsn).Last();
-    }
+                                       !StripedBackupEqualityComparer.Instance.Equals(lastDiff, b));
+
+      var striped = StripedBackupSet.GetStripedBackupSetChain(possibleLogs);
+
+      if (striped.Count() > 1)
+      {
+       return striped.OrderBy(striped => striped.StripedBackups.First().LastLsn).Last().StripedBackups;
+      }
+
+      return possibleLogs;
+  }
 
     private static bool IsValidFilePath(BackupMetadata meta)
     {
--- a/src/BackupChain.cs
+++ b/src/BackupChain.cs
@@ -35,7 +35,7 @@ namespace AgDatabaseMove
       IEnumerable<BackupMetadata> nextLogBackups;
       while((nextLogBackups = NextLogBackup(backups, prevBackup)).Any()) {
         orderedBackups.AddRange(nextLogBackups);
-        prevBackup = orderedBackups.Last();
+        prevBackup = nextLogBackups.First();
       }
 
       _orderedBackups = orderedBackups;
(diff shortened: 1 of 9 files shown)


AdityaNPL (changes requested):
The `OrderBy` clause seems fine. Doesn't look like it will break anything.
But I think it is redundant.

I'm a bit confused - in the first Iteration `nextLog` should not equal 2 diff log backups.
It should only have 1 backup or striped backups of the same backup. (i.e. first & last LSN should be the same)
The `StripedBackupEqualityComparer` takes care of that:
https://github.com/factset/AgDatabaseMove/blob/8111a2221c402a89185489fefe2460ba842566a1/src/BackupMetadata.cs#L16-L17

Which 2 separate log backups are you seeing being added in the same iteration?

mja9:
So, looking at this example data here:
https://github.com/factset/AgDatabaseMove/blob/8111a2221c402a89185489fefe2460ba842566a1/tests/AgDatabaseMove.Unit/BackupOrder.cs#L110-L141

For clarity let's say the elements above are:
Log2
Log1
Diff
 
and bearing in mind the logic in `NextLogBackup` here:
https://github.com/factset/AgDatabaseMove/blob/8111a2221c402a89185489fefe2460ba842566a1/src/BackupChain.cs#L94-L97

Then 
```
Diff.LastLsn = 95000000037000001;
Log1.LastLsn = 95000000037000001;
Log2.LastLsn = 95000000037500001;

: Diff.LastLsn = Log1.LastLsn AND Diff.LastLsn < Log2.LastLsn

Log1.FirstLsn = 95000000036200001;
Log2.FirstLsn = 95000000037000001;

: Diff.LastLsn > Log1.FirstLsn AND Diff.LastLsn = Log2.FirstLsn
```
Since the above is true, both Log1 and Log 2 are of backup type Log, and since none of these backups are stripes of each other (easily observable), both Log1 and Log2 fit the criteria in the `.Where()` clause of `NextLogBackup`.

AdityaNPL:
Ah gotcha, so looks like this happens when a Log backup begins when a diff backup is still completing, but both finish at the same time.
Or at least the LSNs make it look like that's what happened.

I see how the order-by clause will fix this. 
However, it will still get 2 log backups in the same iteration. 
i.e It will get `[..., log 1, log 2]` in the same iteration
(`log2` won't be picked up again cause it is now the last backup which is good)

But can we change the code so that we still fix this bug and only get one backup in each iteration?
Maybe order them by lastLSN in the beginning? Like here before we go into the while loop:
https://github.com/factset/AgDatabaseMove/blob/fix/backupchain/construction/src/BackupChain.cs#L36

I feel that way it's easier to visualise what's going on and debug in the future. Makes it easier when things are consistent 
(i.e. one iteration == one backup being added to the chain)

At least that's my opinion, open to other view points.

tests/AgDatabaseMove.Unit/BackupOrder.cs
```diff
@@ -215,6 +276,17 @@ public void DuplicateFiles()
...
 
+    [Fact]
+    public void SingleServerChainEquality()
+    {
+      var list = GetBackupListSingleServer();
+      var agDb = new Mock<IAgDatabase>();
+      agDb.Setup(db => db.RecentBackups()).Returns(list);
+      var chain = new BackupChain(agDb.Object).OrderedBackups.ToList();
+      list.Reverse();
+      Assert.Equal(list, chain);
```
AdityaNPL, line 287: Have a look at how we are asserting in the other test cases. Would be better to keep this consistent with those.

If this check-function needs changing, that's fine:
`VerifyListIsAValidBackupChain(backupChain.OrderedBackups.ToList());`

tests/AgDatabaseMove.Unit/BackupOrder.cs
```diff
@@ -93,7 +93,68 @@ public static List<BackupMetadata> GetBackupList()
       };
     }
 
-    private static List<BackupMetadata> GetBackupListWithoutLogs()
+  public static List<BackupMetadata> GetBackupListSingleServer()
+  {
```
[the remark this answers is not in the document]
AdityaNPL: ...and the function name for the test-case below will need changing too

AdityaNPL (commented):
Couple of suggestions on the test code changes

alexirion10:
> So, looking at this example data here:
> 
> https://github.com/factset/AgDatabaseMove/blob/8111a2221c402a89185489fefe2460ba842566a1/tests/AgDatabaseMove.Unit/BackupOrder.cs#L110-L141
> 
> For clarity let's say the elements above are: Log2 Log1 Diff
> 
> and bearing in mind the logic in `NextLogBackup` here:
> 
> https://github.com/factset/AgDatabaseMove/blob/8111a2221c402a89185489fefe2460ba842566a1/src/BackupChain.cs#L94-L97
> 
> Then
> 
> ```
> Diff.LastLsn = 95000000037000001;
> Log1.LastLsn = 95000000037000001;
> Log2.LastLsn = 95000000037500001;
> 
> : Diff.LastLsn = Log1.LastLsn AND Diff.LastLsn < Log2.LastLsn
> 
> Log1.FirstLsn = 95000000036200001;
> Log2.FirstLsn = 95000000037000001;
> 
> : Diff.LastLsn > Log1.FirstLsn AND Diff.LastLsn = Log2.FirstLsn
> ```
> 
> Since the above is true, both Log1 and Log 2 are of backup type Log, and since none of these backups are stripes of each other (easily observable), both Log1 and Log2 fit the criteria in the `.Where()` clause of `NextLogBackup`.

So reading over your explanation here.... Log2 shouldn't exists in the backup chain that we put together.  As Aditya mentioned, it looks like the Log got kicked off before the Diff finished, which resulted in both with the same LastLSN.
While `OrderBy` like this could help, the better way to approach this, and how this code was meant to work was by "chaining".  IE the LastLSN must == FirstLSN of the next backup item.
This is the best explanation I have seen http://sqlbak.com/academy/backup-chain

src/BackupChain.cs
```diff
@@ -90,13 +91,30 @@ private static IEnumerable<BackupMetadata> MostRecentFullBackup(IEnumerable<Back
...
+                                       !StripedBackupEqualityComparer.Instance.Equals(lastDiff, b));
+
+      var striped = StripedBackupSet.GetStripedBackupSetChain(possibleLogs);
+
+      if (striped.Count() > 1)
+      {
+       return striped.OrderBy(striped => striped.StripedBackups.First().LastLsn).Last().StripedBackups;
+      }
+
+      return possibleLogs;
```
mja9, line 115: This was the best way I found to fix the issue without larger refactors. Notice how I essentially need to get the stripes from the possible logs in order to verify the true count of backups present. Ideally, the change I outlined [here](https://github.com/factset/AgDatabaseMove/pull/110) would be a better overall fix/change to this class.

alexirion10: hmm.  Looks like it would work but seems messy.  Lets try for the approach in your other PR if that works

alexirion10 (approved):
big changes coming through!  all looks good to me here!

AdityaNPL:
Oh, do you mind changing the PR title to be more informative pls??

PR #110, opened 2022-10-20
refactor(backupchain): group striped backups first, then create the chain

mja9:
While working on the changes in https://github.com/factset/AgDatabaseMove/pull/109, it occurred to me the best way to solve the issues I ran into was to refactor the backup chain construction such that the list of backups obtained from SQL grouped into striped backups first before attempting to build the chain. 

By doing so and by having `StripedBackupSet` inherit from `BackupMetaData`, the refactoring could be kept to a minimal level and much of the added to support striped backups could be removed. I believe @AdityaNPL this is the best way to enforce that a single backup be added to the chain on each iteration (by changing the method signatures to return a single `StripedBackupSet`).

I had to change a couple of the tests since the underlying assumptions/code behavior change as a result of refactoring.

Let me know your thoughts, I sort of let myself get carried away with this one :).


alexirion10:
can you fix the git conflict?

src/BackupChain.cs
```diff
@@ -66,47 +74,39 @@ private static IEnumerable<BackupMetadata> MostRecentFullBackup(IEnumerable<Back
...
-                    .OrderBy(b => b.First().LastLsn).Last();
+      var possibleLogs = backups.Where(b => b.BackupType == BackupFileTools.BackupType.Log &&
+                                       lastBackup.LastLsn >= b.FirstLsn &&
+                                       lastBackup.LastLsn <= b.LastLsn);
+      if (possibleLogs.Count() > 1)
+      {
+        return possibleLogs.OrderBy(b => b.LastLsn).Last();
+      }
+
+      return possibleLogs.SingleOrDefault();
```
AdityaNPL, line 109: This could maybe just be:

```csharp
return possibleLogs.OrderByDescending(b => b.LastLsn).FirstOrDefault();
```

src/BackupMetadata.cs
```diff
@@ -89,15 +89,29 @@ public object Clone()
...
+    {
+     CheckpointLsn = firstStripe.CheckpointLsn;
+     DatabaseBackupLsn = firstStripe.DatabaseBackupLsn;
+     DatabaseName = firstStripe.DatabaseName;
+     FirstLsn = firstStripe.FirstLsn;
+     LastLsn = firstStripe.LastLsn;
+     PhysicalDeviceName = null;
+     ServerName = firstStripe.ServerName;
+     StartTime = firstStripe.StartTime; // or is null more appropriate?
+     BackupType = firstStripe.BackupType;
```
AdityaNPL, line 112: There is a more OOP way to do this:
We create a constructor in `BackupMetadata` that takes a `BackupMetadata` obj (like a copy constructor).
Then here, in the `StripedBackupSet` constructor, just call `base(stripedBackups.First())`

That way the properties are initialised in the class that declares them

src/BackupMetadata.cs
```diff
@@ -89,15 +89,29 @@ public object Clone()
...
     }
 
+    private void SetMetadata(BackupMetadata firstStripe)
+    {
+     CheckpointLsn = firstStripe.CheckpointLsn;
+     DatabaseBackupLsn = firstStripe.DatabaseBackupLsn;
+     DatabaseName = firstStripe.DatabaseName;
+     FirstLsn = firstStripe.FirstLsn;
+     LastLsn = firstStripe.LastLsn;
+     PhysicalDeviceName = null;
```
AdityaNPL, line 109: I'm on the fence about setting `null` for `PhysicalDeviceName`.

Have debated with myself for a good 5 mins and come to the following conclusion 😅:
"the `null` here signifies that there isn't just **one** physical device name" 

This sounds ok to me, not sure if there is a better way to do this

src/BackupMetadata.cs
```diff
@@ -89,15 +89,29 @@ public object Clone()
...
+     ServerName = firstStripe.ServerName;
+     StartTime = firstStripe.StartTime; // or is null more appropriate?
```
AdityaNPL, line 111: start time will be the same for all the stripes, don't need null here

src/BackupMetadata.cs
AdityaNPL: Will still need to set `PhysicalDeviceName=null` in the `StripedBackupSet` constructor tho - but that makes sense, it's relevant to this class

AdityaNPL (changes requested):
I like the idea of grouping the raw backups into their stripes first and then forming the chain.
That way we can treat each group of striped backups as one entity.

Thanks for looking into this and making the required changes.

Have a few code syntax suggestions below and small code design improvements.
But apart from that, looks like the logic is sound and seems like it should work. 👍🏽
But to know for sure, do you mind running the integration tests and E-2-E (with and without striped backups) pls?

(and yes, pls fix the conflict)

src/BackupChain.cs
```diff
@@ -8,35 +8,44 @@ namespace AgDatabaseMove
...
 
-      var orderedBackups = MostRecentFullBackup(backups).ToList();
-      orderedBackups.AddRange(MostRecentDiffBackup(backups, orderedBackups.First()));
-      orderedBackups.AddRange(FirstLogBackupInChain(backups, orderedBackups.Last()));
+      var orderedBackups = new List<StripedBackupSet> { MostRecentFullBackup(stripedBackups) };
+      var diff = MostRecentDiffBackup(stripedBackups, orderedBackups.First());
+      if (diff != null)
+      {
+       orderedBackups.Add(diff);
+      }
```
AdityaNPL, line 37: Could be inline...but no strong preference from my side

src/BackupChain.cs
```diff
@@ -66,54 +75,35 @@ private static IEnumerable<BackupMetadata> MostRecentFullBackup(IEnumerable<Back
...
-      if(!diffBackupsOrdered.Any())
-        return new List<BackupMetadata>();
-
-      var targetLastLsn = diffBackupsOrdered.First().LastLsn;
-      // get all the stripes of this backup
-      return diffBackupsOrdered.Where(diffBackup => diffBackup.LastLsn == targetLastLsn);
+      return backups.Where(b => b.BackupType == BackupFileTools.BackupType.Diff &&
+                                b.DatabaseBackupLsn == lastFullBackup.CheckpointLsn)
+                    .OrderByDescending(b => b.LastLsn)
+                    .FirstOrDefault();
```
AdityaNPL, line 88: I like what you did with the others - removing the redundant `Where()` and putting the predicate logic in the `FirstOrDefault()`.
Let's keep that consistent, do you mind changing that here as well pls?

src/BackupChain.cs
```diff
@@ -66,54 +75,35 @@ private static IEnumerable<BackupMetadata> MostRecentFullBackup(IEnumerable<Back
...
+                                       lastBackup.LastLsn <= b.LastLsn).ToList();
 
-      if (striped.Count() > 1)
-      {
-       return striped.OrderBy(striped => striped.StripedBackups.First().LastLsn).Last().StripedBackups;
-      }
-
-      return possibleLogs;
-  }
+      return possibleLogs.OrderByDescending(b => b.LastLsn).FirstOrDefault();
```
AdityaNPL, line 105: Re-reading this, I think it can be combined with the above LINQ.
There isn't anything that complicated we are doing in this second LINQ.

Combining them also means we are removing the `ToList()` so it stays as an IEnumerable until the `FirstOrDefault`.

(Also once again, `Where()` can be removed)

tests/AgDatabaseMove.Unit/BackupOrder.cs
```diff
@@ -205,14 +205,15 @@ private static List<BackupMetadata> GetBackupListWithoutFull()
...
+      Assert.True(backupChain.Any());
       bool foundFull, foundDiff, foundLog;
       foundFull = foundDiff = foundLog = false;
-      BackupMetadata full = null;
-      BackupMetadata lastBackup = null;
-
-      BackupMetadata currentBackup;
+      StripedBackupSet full = null;
+      StripedBackupSet lastBackup = null;
+
```
AdityaNPL, line 215: Extra space

src/BackupChain.cs
```diff
@@ -55,9 +64,9 @@ private BackupChain(IList<BackupMetadata> recentBackups)
     /// <summary>
     ///   Backups ordered to have a full restore chain.
     /// </summary>
-    public IEnumerable<BackupMetadata> OrderedBackups => _orderedBackups;
+    public IEnumerable<StripedBackupSet> OrderedBackups => _orderedBackups;
 
-    private static IEnumerable<BackupMetadata> MostRecentFullBackup(IEnumerable<BackupMetadata> backups)
+    private static StripedBackupSet MostRecentFullBackup(IEnumerable<StripedBackupSet> backups)
```
AdityaNPL, line 69: Using the variable name `backups` is a bit misleading here, especially since we are using `stripedBackups` in the calling function.
Can you please change it to `stripedBackups` here and in the other functions for the similar data?

src/BackupChain.cs
```diff
@@ -66,54 +75,35 @@ private static IEnumerable<BackupMetadata> MostRecentFullBackup(IEnumerable<Back
       if(!fullBackupsOrdered.Any())
         throw new BackupChainException("Could not find any full backups");
 
-      var targetCheckpointLsn = fullBackupsOrdered.First().CheckpointLsn;
       // get all the stripes of this backup
```
AdityaNPL, line 78: Don't think we need this comment anymore

AdityaNPL (commented):
Looking much better, thanks!
A few more...

slin5299 (commented):
I also like this approach more and I think this is pretty much good to go after Adi's suggestions. Any updates on what testing you've done?

mja9:
> I also like this approach more and I think this is pretty much good to go after Adi's suggestions. Any updates on what testing you've done?

The unit tests all work and so do most of the integration tests -- I'm getting a weird null exception in the restore tests about a `sleepDurationProvider` I'm still trying to find.

src/BackupChain.cs
mja9: It could be combined, but I personally prefer keeping the methods separate. In `NextLogBackup` the query should only match a single backup or none at all, which can be enforced through the use of `SingleOrDefault()`. The rule in `FirstLogInChain` isn't as strict since it could potentially lead to the edge-case encountered in https://github.com/factset/AgDatabaseMove/pull/109.

AdityaNPL: > but I personally prefer keeping the methods separate

ah, no - looks like there has been a misunderstanding here.

I'm not referring to combining the functions. I'm taking about combining the LINQ in this `FirstLogInChain()` function.
...but looks like you have done that now 👍🏽

AdityaNPL:
> > I also like this approach more and I think this is pretty much good to go after Adi's suggestions. Any updates on what testing you've done?
> 
> The unit tests all work and so do most of the integration tests -- I'm getting a weird null exception in the restore tests about a `sleepDurationProvider` I'm still trying to find.

If you are still having issues with this, can you pls share more details here?
I can try to have a look from my side as well.

mja9:
> > > I also like this approach more and I think this is pretty much good to go after Adi's suggestions. Any updates on what testing you've done?
> > 
> > 
> > The unit tests all work and so do most of the integration tests -- I'm getting a weird null exception in the restore tests about a `sleepDurationProvider` I'm still trying to find.
> 
> If you are still having issues with this, can you pls share more details here? I can try to have a look from my side as well.

All tests are passing now, integration included.

src/BackupChain.cs
```diff
@@ -8,35 +8,38 @@ namespace AgDatabaseMove
...
 
-      var orderedBackups = MostRecentFullBackup(backups).ToList();
-      orderedBackups.AddRange(MostRecentDiffBackup(backups, orderedBackups.First()));
-      orderedBackups.AddRange(FirstLogBackupInChain(backups, orderedBackups.Last()));
+      var orderedBackups = new List<StripedBackupSet> { MostRecentFullBackup(stripedBackups) };
+      var diff = MostRecentDiffBackup(stripedBackups, orderedBackups.First());
+      if (diff != null) { orderedBackups.Add(diff); }
+
+      var nextLog = FirstLogInChain(stripedBackups, orderedBackups.Last());
+      if (nextLog != null) { orderedBackups.Add(nextLog); }
```
AdityaNPL, line 37: ...
else return

(can return early if there are no log backups to add)

AdityaNPL (approved):
LGTM!
(Just noticed one small improvement we can make below)

Since the integration tests are passing (and it has restore tests) we should be fine.

Thanks for making the changes here!

src/BackupChain.cs
AdityaNPL: That works too, thanks!

alexirion10:
I'll take another look over this today

src/BackupMetadata.cs
```diff
@@ -82,23 +82,39 @@ public class BackupMetadata : ICloneable
...
     public object Clone()
     {
       return MemberwiseClone();
     }
   }
 
-  public class StripedBackupSet
+  public class StripedBackupSet : BackupMetadata
   {
     public IEnumerable<BackupMetadata> StripedBackups { get; private set; }
```
alexirion10, line 109: Can you rename `StripedBackupSet` to `StripedBackup`.  There is no 'set' collection here.  It's 1 entity with a list of different filenames.

src/BackupMetadata.cs
```diff
@@ -82,23 +82,39 @@ public class BackupMetadata : ICloneable
...
 
-    private StripedBackupSet(IEnumerable<BackupMetadata> stripedBackups)
+    private StripedBackupSet(IEnumerable<BackupMetadata> stripedBackups) : base(stripedBackups.First())
     {
       StripedBackups = stripedBackups;
+      PhysicalDeviceName = null;
```
alexirion10, line 114: Some discussion is needed here:
The only thing that differs between each "stripe" is the `PhysicalDeviceName`.  Why are you setting it to null here?  That is a confusing usage of the inherited value.  I think you can do better by adding some class hierarchy to this:

```C#
public abstract class BackupMetadata
{ /*all of the fields except PhysicalDeviceName*/}

public class SingleBackupMetadata : BackupMetadata
{ PhysicalDeviceName = backup.PhysicalDeviceName; }

public class StripedBackup : BackupMetadata
{  StripedBackups = stripedBackups; }
```

alexirion10 (commented):
Looks great!  Two small things below before shipping this

src/BackupMetadata.cs
```diff
@@ -84,41 +83,44 @@ public class BackupMetadata : ICloneable
...
 
-    private StripedBackupSet(IEnumerable<BackupMetadata> stripedBackups) : base(stripedBackups.First())
+    private StripedBackup(IEnumerable<SingleBackup> stripedBackups) : base(stripedBackups.First())
     {
       StripedBackups = stripedBackups;
-      PhysicalDeviceName = null;
     }
 
-    internal static IEnumerable<StripedBackupSet> GetStripedBackupSetChain(IEnumerable<BackupMetadata> backups)
+    internal static IEnumerable<StripedBackup> GetStripedBackupSetChain(IEnumerable<SingleBackup> backups)
```
alexirion10, line 119: `GetStripedBackupSetChain` --> `GetStripedBackupChain`

src/BackupMetadata.cs
```diff
@@ -84,41 +83,44 @@ public class BackupMetadata : ICloneable
...
+  public class SingleBackup : BackupMetadata
+  {
+    public string PhysicalDeviceName { get; set; }
   }
 
-  public class StripedBackupSet : BackupMetadata
+  public class StripedBackup : BackupMetadata
   {
-    public IEnumerable<BackupMetadata> StripedBackups { get; private set; }
+    public IEnumerable<SingleBackup> StripedBackups { get; }
```
alexirion10, line 112: I could see this being a `List<string>` which only contains the PhysicalDeviceName's .... but ok this works fine too.
In the constructor, you would convert by doing `stripedBackups.Select(i=> i.PhysicalDeviceName);`
If it involves more refactoring, don't worry about it.  If it's simple, I think it'll simplify the meaning here

alexirion10 (approved):
🚀

PR #113, opened 2022-11-23
feat(Restore): single backup

mja9:
Created an overload for `Restore` that takes in a full `SingleBackup`. This can be used to restore databases that can be effectively recovered under the simple recovery model. Added a couple integration tests and verified they passed.

### DIFF
--- a/src/SmoFacade/Server.cs
+++ b/src/SmoFacade/Server.cs
@@ -168,10 +168,34 @@ namespace AgDatabaseMove.SmoFacade
       return database == null ? null : new Database(database, this);
     }
 
+    public void Restore(SingleBackup fullBackup, string databaseName,
+      Func<int, TimeSpan> retryDurationProvider,
+      Func<string, string> fileRelocation = null)
+    {
+      var retryPolicy = Policy
+        .Handle<ExecutionFailureException>(e => e.InnerException != null
+                                                && e.InnerException is SqlException
+                                                && e.InnerException.Message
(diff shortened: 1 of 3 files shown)


alexirion10 (approved):
🦃

slin5299 (approved):
lgtm!

src/SmoFacade/Server.cs
```diff
@@ -168,10 +168,34 @@ public Database Database(string dbName)
...
+    public void Restore(SingleBackup fullBackup, string databaseName,
+      Func<int, TimeSpan> retryDurationProvider,
+      Func<string, string> fileRelocation = null)
+    {
+      var retryPolicy = Policy
+        .Handle<ExecutionFailureException>(e => e.InnerException != null
+                                                && e.InnerException is SqlException
+                                                && e.InnerException.Message
+                                                  .Contains("The process cannot access the file because it is being used by another process"))
+        .WaitAndRetry(10, retryDurationProvider);
```
AdityaNPL, line 180: `retryPolicy` can be placed into a private variable and then re-used by both restore functions.
Maybe rename to `_restoreRetryPolicy`?

AdityaNPL: looks like it needs `retryDurationProvider` (maybe a private utility method then)?

AdityaNPL (approved):
Makes sense, code LGTM!
Tests are sound too.

Just one small suggestion below

PR #117, opened 2023-08-29
feat(IListener): visibility

mja9:
I have a need for the internals of `IListener` to be made public. Specifically, I'm trying to avoid having to re-write code to get the names of/connections to replicas in an availability group.

The concrete implementation `Listener` is still internal and therefore can't be directly instantiated... I believe that should be sufficient protection.

### DIFF
--- a/src/AgDatabase.cs
+++ b/src/AgDatabase.cs
@@ -62,7 +62,6 @@ namespace AgDatabaseMove
   public class AgDatabase : IDisposable, IAgDatabase
   {
     private readonly string _backupPathSqlQuery;
-    internal readonly IListener _listener;
 
     /// <summary>
     ///   A constructor that uses a config object for more options.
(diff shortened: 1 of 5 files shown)


src/AgDatabase.cs
```diff
@@ -79,6 +80,8 @@ public AgDatabase(DatabaseConfig dbConfig)
 
     public int ServerRemainingDiskMb => _listener.Primary.RemainingDiskMb();
 
+    public IListener Listener => _listener;
```
alexirion10, line 83: Instead of this accessor to a private member variable, it's a bit more clear to make this a property
https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/using-properties#the-get-accessor

The [auto implemented property](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/auto-implemented-properties) example that you'd use here:
`public IListener Listener { get; }`

alexirion10 (commented):
looks good, the CI build needs fixin'

mja9:
> looks good, the CI build needs fixin'

Looking into it, but I'm having trouble locating the issue causing the failure.

alexirion10 (approved):

PR #118, opened 2023-09-12
chore(Polly): 7.2.4

mja9:
Update Polly package `7.2.3 --> 7.2.4`


alexirion10 (changes requested):
Seems like 2 very different changes grouped together here.  Can you split the pipeline YML fix off from the Nuget package update(s)?

alexirion10 (approved):

PR #120, opened 2023-09-25
fix(syslogins): password NULL

mja9:
Fixes the issue described in https://www.tarynpivots.com/post/system-view-gotcha-with-sql-server-2019/.

In SqlServer 2019, the _sys.syslogins_ view returns `NULL` for the `password` field. _sys.sql_logins_ should be used instead.
Should continue to work for older version of SqlServer as well.

### DIFF
--- a/src/SmoFacade/Login.cs
+++ b/src/SmoFacade/Login.cs
@@ -54,7 +54,9 @@ namespace AgDatabaseMove.SmoFacade
     /// <returns>A string containing the hex representation of the password hash.</returns>
     public string PasswordHash()
     {
-      var sql = "SELECT password_hash as passwordHash FROM sys.sql_logins WHERE name = @loginName";
+      var sqlVersion = _server._server.Version;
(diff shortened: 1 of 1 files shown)


alexirion10 (commented):
Will this `password_hash` work when adding it to the new location?  
Have you tested a database copy/move using this?

mja9:
> Will this `password_hash` work when adding it to the new location? Have you tested a database copy/move using this?

Yes, I've verified this works by copying over logins between servers and connecting using those logins.

alexirion10:
Also - will this new logic work for SQLServer 2017 or earlier?  This might need logic in here dependent on the server version

mja9:
> Also - will this new logic work for SQLServer 2017 or earlier? This might need logic in here dependent on the server version

This is compatible back to Sql 2016. I'm looking into whether there should be version dependent logic here or not. I think it should depend on the earliest Sql version compatible with the Smo library version being used.

mja9:
This should now be compatible with older versions of SqlServer as well.

alexirion10 (approved):
Looks great!  Thanks for figuring out the changed behaviour between MSSQL versions

PR #123, opened 2026-04-01
feat(backup): optional file name

mja9:
This change adds an option to configure backup file naming. The naming convention was previously hardcoded in `Server.cs`. This provides the ability to customize that naming convention with an optional delegate which takes `databaseName` and `backupType` since these are commonly used in backup names.


jcostamartin (approved):
LGTM

